PR #3, opened 2018-09-18
Fix undefined method `content_type' for nil:NilClass

schweigert:
Hi,

I had some problems using this gem, because blob does not exist in some cases.
I'm fixing this problem.


igorkasyanchuk:
merged, new version of gem release (please try it)
also I've mentioned you here: https://github.com/igorkasyanchuk/active_storage_validations/blob/master/README.md#contributors-big-thank-you
thanks!

PR #4, opened 2018-09-19
content_size validation added

reckerswartz:
## content_size validation added for 
:less_than, 
:less_than_or_equal_to, 
:greater_than, 
:greater_than_or_equal_to, 
:between

*writing test pending!

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ class Project < ApplicationRecord
 
   validates :title, presence: true
 
-  validates :preview, attached: true, content_size: { greater_than: 100.megabytes , message: 'is not given between size' }
+  validates :preview, attached: true, content_size: { less_than: 100.megabytes , message: 'is not given between size' }
   validates :attachment, attached: true, content_type: { in: 'application/pdf', message: 'is not a PDF' }
 end
 ```
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ class Project < ApplicationRecord
 
   validates :title, presence: true
 
-  validates :preview, attached: true
+  validates :preview, attached: true, content_size: { greater_than: 100.megabytes , message: 'is not given between size' }
   validates :attachment, attached: true, content_type: { in: 'application/pdf', message: 'is not a PDF' }
 end
 ```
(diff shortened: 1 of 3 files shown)


lib/active_storage_validations.rb
```diff
@@ -44,6 +44,50 @@ def content_type_valid?(file)
...
+          file_size >= options[:less_than_or_equal_to]
+        when options[:greater_than].present?
+          file_size < options[:greater_than]
+        when options[:greater_than_or_equal_to].present?
+          file_size <= options[:greater_than_or_equal_to]
+      end 
+    end
+
+    def human_size(size)
+      ActiveSupport::NumberHelper.number_to_human_size(size)
```
igorkasyanchuk: I think you can do something like

delegate :number_to_human_size, to: ActiveSupport::NumberHelper

and remove this method

README.md
```diff
@@ -29,7 +29,7 @@ class Project < ApplicationRecord
 
   validates :title, presence: true
 
-  validates :preview, attached: true
+  validates :preview, attached: true, content_size: { less_than: 100.megabytes , message: 'is not given between size' }
```
igorkasyanchuk: maybe rename content_size to "size" or "file_size"
since it's more about file

igorkasyanchuk:
@reckerswartz very good start, please check few comments and yes, at least a simple tests are needed

README.md
reckerswartz: @igorkasyanchuk request to change it to active_storage_size prevent conflict with other 3party gems. please suggest.

igorkasyanchuk: this is for "size" and "file_size"? who is using such validations? Maybe file_size is available?

igorkasyanchuk: @reckerswartz do you have any other comments? this PR looks like 95% ready, if you can make fixes I'll merge it.

So I see 2 things which I would like to change:
- name of validator (find a better name if possible)
- use number_to_human_size (see my other comment in PR)

reckerswartz: @igorkasyanchuk sorry for delay!
updated as per requested but still test are pending :smile:

igorkasyanchuk:
merged, I'll make a new release soon

igorkasyanchuk:
@reckerswartz version 0.5 just released. thanks

PR #5, opened 2018-09-22
Add better error message when content type is invalid

tleneveu:


lib/active_storage_validations.rb
```diff
@@ -20,8 +20,8 @@ def validate_each(record, attribute, value)
       files = Array.wrap(files)
 
       files.each do |file|
-        unless content_type_valid?(file) 
-          record.errors.add(attribute, options[:message].presence || :invalid)
+        unless content_type_valid?(file)
+          record.errors.add(attribute, options[:message].presence || "does not have an authorized content type, authorized content types : #{types}")
```
igorkasyanchuk: @tleneveu  this is something that we can discuss. I think ideally gem should be without hard-coded language. Maybe with a new key and ability to localize it. What do you think? This is why I put :invalid, since I think with different format it would be translated automatically

igorkasyanchuk: also if only one type is specified it should output string like "authorized content type : #{types}" (without "s")
and maybe not showed as array, but in more human readable format (maybe joined with ",")

tleneveu: Effectively

I think it would be better to add internationalization for providing multi-language support

igorkasyanchuk:
@tleneveu does it all work well? if yes - I'll merge

tleneveu:
@igorkasyanchuk I think it would be better to rename types method to authorized_types

What do you think?

igorkasyanchuk:
Code is good  but I don't like the message http://prntscr.com/kxq5a4 (two identical parts)
maybe something more common to use?
let's maybe use еру same message as carrierwave does https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/locale/en.yml#L7

also can you add a sample of locale file what users can edit and change https://github.com/carrierwaveuploader/carrierwave/wiki/How-to%3A-Validate-uploads-with-Active-Record like here

so 2 things:
- message similar to carrierwave
- documentation about i18n

GBH:
One thing I'd like to see is what content type is actually there. Like `invalid file content_type "foo/bar"`. More useful than listing all allowed content types (I might have like a hundred).

igorkasyanchuk:
@GBH what full message you propose?

GBH:
I think I'd be happy to see "Invalid file type: %{content_type}". End-user gets a hint that they are uploading word document instead of image (hopefully) and developer can check if file's content type is properly defined in the validator.

igorkasyanchuk:
I'm a big fan of using positive messages and more clear. I think better to show what is allowed. Since user can try more and more until he get an idea what is allowed.

GBH:
Doesn't work that well with large whitelist. Also you're assuming that end-user knows what `application/vnd.openxmlformats-officedocument.wordprocessingml.document` means.

igorkasyanchuk:
okay, but what if user is trying to upload XLSX but only PNG are allowed:
by your logic error message will be:

"Content type application/vnd.openxmlformats-officedocument.wordprocessingml.document is not allowed"


I think maybe the easiest is to show just "<field> is invalid". A user always can put a custom error message if needed.

GBH:
Yeah. Custom message is the best. Just thinking that the default should be what I suggested so it's possible to debug when developing.

igorkasyanchuk:
taking into consideration all previous comments, as much as I think about this PR and what is the best option for the error message I think that may be the ideal case is to leave as it's and use custom erro message where user can express what he wants

tleneveu:
I propose this error message : Image has an invalid content type

It remains simple and understandable.

It would be better to leave variables for the user to customize, a variable containing the content type of the send file and a variable containing the list of authorized content types.

tleneveu:
@igorkasyanchuk What do you think of the changes brought ?

igorkasyanchuk:
new release active_storage_validations-0.4.0.gem
please try it
thanks

PR #10, opened 2018-10-04
Remove Pry from Gemfile as it is already present in gemspec

reckerswartz:
Repeat of #1 
Pry is added as a development dependency in active_storage_validations.gemspec, as well as in the Gemfile, which leads to the following command line warning message upon running bundle install
```
Your Gemfile lists the gem pry (>= 0) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of one of them later.
```
This Pull Request removes Pry from the Gemfile.
Gemfile.lock has not changed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,3 @@ gemspec
 
 # To use a debugger
 # gem 'byebug', group: [:development, :test]
-
-group :development, :test do
-  gem 'pry'
-end


igorkasyanchuk:
merged, thanks

igorkasyanchuk:
client side validations are fine, but I think they should be a part of that gem.

reckerswartz:
@igorkasyanchuk I am talking about new gem for client-side-validation with rails form_with support, as form_for is soft deprecated.

igorkasyanchuk:
ohh I see. Yes, this is a nice idea. I've also a few ideas, will share later.

Issue #12, opened 2018-10-17
image upload size


dbbooke:
Is there any validations for how many images a user can uploaded? For example, each post can post up to 24 images?

reckerswartz:
Nope, never thought of that.

reckerswartz:
@igorkasyanchuk do you want this to be handled in new validation class or i update it in size validation class?

igorkasyanchuk:
Hi, yes, new validation class. Something like

validates ... attachments_count: { max: 15 OR min: 10  }

ivanelrey:
Hi @igorkasyanchuk can you please review the pull request and tell me if any changes needed, thanks?

igorkasyanchuk:
yes, already commented. very minor fix is needed not related to code

PR #16, opened 2019-01-27
Validation of total attachments.

ivanelrey:
Issue: #12 
Example:

  validates :files, limit { min: 1, max: 5 }

  An error will occur if total number of attached files are below 1 or greater that 5.

Error message:
  limit_out_of_range: "total number is out of range"

Also max and min can be used in the error message.
  limit_out_of_range: "total number should be between %{min} and %{max}"

 On branch add-limit-validation
 Changes to be committed:
	modified:   ../config/locales/en.yml
	modified:   ../lib/active_storage_validations.rb
	modified:   active_storage_validations_test.rb
	new file:   dummy/.byebug_history
	modified:   dummy/app/models/project.rb

 Untracked files:
	../lib/active_storage_validations/limit_validator.rb


test/dummy/.byebug_history
```diff
@@ -0,0 +1,40 @@
+continue
```
igorkasyanchuk: please add to gitignore

igorkasyanchuk:
I've only one comment.
Please fix and I'll merge, looks very good and useful

igorkasyanchuk:
@ivanelrey  CC

ivanelrey:
@igorkasyanchuk Done. Thanks a lot.

igorkasyanchuk:
new version just released. thanks

PR #28, opened 2019-04-16
Declare image_processing optional dependency

arr-dev:
Loaded explicitly in dimension validation,
since it's only required there.

My app uploads only txt files so there is no need for `image_processing`  gem in such apps.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -55,14 +55,10 @@ GEM
     concurrent-ruby (1.1.5)
     crass (1.0.4)
     erubi (1.8.0)
-    ffi (1.10.0)
     globalid (0.4.2)
       activesupport (>= 4.2.0)
     i18n (1.6.0)
       concurrent-ruby (~> 1.0)
-    image_processing (1.9.0)
-      mini_magick (>= 4.9.3, < 5)
-      ruby-vips (>= 2.0.13, < 3)
     jaro_winkler (1.5.2)
     loofah (2.2.3)
       crass (~> 1.0.2)
@@ -125,8 +121,6 @@ GEM
       ruby-progressbar (~> 1.7)
       unicode-display_width (~> 1.4.0)
     ruby-progressbar (1.10.0)
-    ruby-vips (2.0.13)
-      ffi (~> 1.9)
     sprockets (3.7.2)
       concurrent-ruby (~> 1.0)
       rack (> 1, < 3)
@@ -149,7 +143,7 @@ PLATFORMS
 
 DEPENDENCIES
   active_storage_validations!
-  image_processing (>= 1.8)
+  mini_magick (>= 4.9)
   pry
   rubocop
   sqlite3
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,6 @@ PATH
   remote: .
   specs:
     active_storage_validations (0.7.0)
-      image_processing (>= 1.8)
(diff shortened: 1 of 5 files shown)


igorkasyanchuk:
I'm ok with this change, let's check with the author of original change @phlegx 

Also please improve readme, this section http://prntscr.com/nco2dz (add image_processing gem with optional comment). So it will be more clear for user.

arr-dev:
Thanks, hope this is what you meant.

phlegx:
Here the code of `ActiveStorage` where dimension `width` and `height` is retrieved from image: [active_storage/analyzer/image_analyzer.rb](https://github.com/rails/rails/blob/v6.0.0.beta3/activestorage/lib/active_storage/analyzer/image_analyzer.rb)

And here is explained why using `image_processing` gem: [edgeguides.rubyonrails.org](https://edgeguides.rubyonrails.org/active_storage_overview.html#transforming-images)

So, we can remove it completely and add `require 'mini_magick'` for check.

- `lib/active_storage_validations/dimension_validator.rb`
- [rubygems.org/gems/mini_magick](https://rubygems.org/gems/mini_magick) -> 4.9.3 is the latest version
- Dimension detection requires `mini_magick` or `image_processing` gem.

arr-dev:
Right, `mini_magick` is a dependency of `image_processing`, but it alone should be enough.
I'll update the code for `mini_magick`

igorkasyanchuk:
@phlegx @arr-dev is it ready for merge?

arr-dev:
I believe it is, let's see what @phlegx says

phlegx:
Looks good and can be merged! 

We need to keep an eye on this method in future: [active_storage/analyzer/image_analyzer.rb#L34](https://github.com/rails/rails/blob/v6.0.0.beta3/activestorage/lib/active_storage/analyzer/image_analyzer.rb#L34). It could be changed using the general `image_processing` gem where a developer can choose between **Imagemagick** and **Vips** to get the height and width of an image.

`ActiveStorage` v.6.x has already documented an example in [activestorage/app/models/active_storage/variation.rb](https://github.com/rails/rails/blob/v6.0.0.beta3/activestorage/app/models/active_storage/variation.rb#L9) where `resize_to_limit` (an `image_processing` parameter) is used to create a variation. Also **Vips** has a method to get [width and height](https://www.rubydoc.info/gems/ruby-vips/Vips%2FImage:size). So, most likely also the [active_storage/analyzer/image_analyzer.rb#L34](https://github.com/rails/rails/blob/v6.0.0.beta3/activestorage/lib/active_storage/analyzer/image_analyzer.rb) could change using the general `image_processing` wrapper.

reckerswartz:
I think most of are going to love **VIPS** and will change to it over **Imagemagick**. Even I am using VIPS in production, it is better to leave the dependency image_processing in Gemspec and mark required true only for those modules which are going to use it so that it will doesn't effect anyhow those who don't want to use it.
suggestion @igorkasyanchuk @phlegx

igorkasyanchuk:
try to use active_storage_validations-0.7.1.gem

thank you ALL for contribution

phlegx:
Master code of ActiveStorage on [active_storage/analyzer/image_analyzer.rb#L36](https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/analyzer/image_analyzer.rb#L36) uses **Imagemagick**. So, we leave it like actual PR of @arr-dev.

Issue #31, opened 2019-06-05
Rails 6.0.0.rc1:  ActiveStorage::FileNotFoundError on model save


oiuzikov:
The following declaration throws `ActiveStorage::FileNotFoundError` in Rails 6.0.0.rc1 when I try to save a model:

```ruby
validates :image, attached: true, content_type: IMAGE_CONTENT_TYPE,
  dimension: { width: IMAGE_WIDTH, height: IMAGE_HEIGHT }
```

with the following cause:

`
Errno::ENOENT: No such file or directory @ rb_sysopen - /home/oleksandri/Code/marketplace/storage/08/3z/083z1pphjimzuwty55kqvb7t1hhj 
`

Is it supported in Rails 6?

Thank you for your valuable work

igorkasyanchuk:
Don't know. Maybe, we don't have specs for different versions. Please post full stack trace to see where is error happens

oiuzikov:
Thank you for quick reply. Here is the trace you asked for.

I also did a quick research and seems like ActiveStorage doesn't save the attachment to the file system before the model is saved which in turn breaks the validation flow...

Extracted source (around line 19): 
```ruby
def create
  @design = Design.new design_params
  if @design.save # <= line 19
    redirect_to user_designs_path
  else
    render :new
```

Exception causes:
```
Errno::ENOENT: No such file or directory @ rb_sysopen - /home/oleksandri/Code/marketplace/storage/8c/ul/8culuojs36e2ol6q5a49g87s48a1
```
Request:
```
{"utf8"=>"✓",
 "authenticity_token"=>"2YS2nziC4UPNLXqipa2VPy73Hr2RXaRcyI7CXyCBLqV/2Q9ZtYLTeYhFzP4ifhiySShLtIR2xvMzNfLY7/M/0A==",
 "design"=>
  {"name"=>"New Design 1",
   "image"=>
    #<ActionDispatch::Http::UploadedFile:0x00007f64643857f0
     @content_type="image/png",
     @headers="Content-Disposition: form-data; name=\"design[image]\"; filename=\"small.png\"\r\n" + "Content-Type: image/png\r\n",
     @original_filename="small.png",
     @tempfile=#<File:/tmp/RackMultipart20190605-31557-1gs32cd.png>>},
 "button"=>""}
```

Full trace:
```
activestorage (6.0.0.rc1) lib/active_storage/service/disk_service.rb:136:in `rescue in stream'
[...]

PR #36, opened 2019-07-29
Add validation error for when image metadata is missing width and/or height

danderozier:
This is a fix for the issue I mentioned here: https://github.com/igorkasyanchuk/active_storage_validations/issues/31#issuecomment-515169356

If width/height keys are missing from the attachment metadata, it indicates that the attachment isn't an image, so instead of raising a StandardError, the validator just adds a validation error to the record instead.

This change affects both dimension and aspect ratio validators.


danderozier:
One other note: to get the tests running, I had to comment out a couple of lines in active_storage_validations_test.rb. The `save!` call on an invalid record was throwing an error and causing the tests to hang.

igorkasyanchuk:
@danderozier I've fixed and improved specs.

I've noticed one strange thing, when we have a PDF file it's actually receiving width/height attributes, so content type is also needed in this case just for your information.

with HTML file for example when I'm trying to attach all works well.

danderozier:
Aha, that makes sense. Thanks for your work on this project!

Issue #6, opened 2018-09-24
Content type validation should handle regex expression


GBH:
`validates :image, content_type: /\Aimage\/.*\z/` Should  be a valid expression

ivanelrey:
@igorkasyanchuk what you think about this issue? Should this be implemented?

igorkasyanchuk:
Probably, if you want to implement it could be useful for some people.

igorkasyanchuk:
Implementerd

PR #38, opened 2019-08-07
validates with content type regex

Uysim:
Answer to the issues #6 which I see it opened since last year


igorkasyanchuk:
Merged, released: active_storage_validations-0.8.2.gem
Thanks!

PR #45, opened 2019-10-20
fix limit validation for rails 6

UICJohn:
Looks like the activestorage in rails 5 cannot prevent blob being created even if the [validations failed ](https://github.com/rails/rails/issues/31656).
Therefore, I test the tweak in rails 6 only.  If anything wrong about the tweak, please tell me. Thanks.


igorkasyanchuk:
Thanks!

igorkasyanchuk:
@UICJohn why you have added "uniq"?

UICJohn:
I tried to uniq the blob files while validation. The changes of active_stroage 'attach' method cause duplicate attachables. That is why I added 'uniq' to validation.

igorkasyanchuk:
I've removed it, because what if user wants to add the same files

UICJohn:
Well, I think it will be fine to use `uniq` for validation.

In my sight, It will have two cases for user adding the same files:

- Batch uploads
`uniq` compares two activerecord objects with `eql?` which according to rails document [eql?](https://api.rubyonrails.org/v6.0.0/classes/ActiveRecord/Core.html#method-i-3D-3D), it won't uniq two new records since the comparison will return false. I will add another test to make sure the validation works in this case

- Add same files one by one
The reason that there is no necessary to worry about in this case is the same as above. I've added the test for this case.

igorkasyanchuk:
OK, but what problem "uniq" is solving? why it's needed?

UICJohn:
As I mentioned before, the changes of the `attach` method in active_storage v6.0 cause duplicate files. For example, when uploading multiple files one by one, the attachable become something like [file_a, file_b, file_a.....], what I wanna do is uniq the attachable and make it like [file_a, file_b.....]. There is a way to reproduce the issue, try print files in `LimitValidator` line 16 and run the test (line 77-80).

igorkasyanchuk:
@UICJohn  hmm, OK, I've added it back. But one moment doesn't it mean it should be also added for other validators?

UICJohn:
I used content and limit validators in my current project and the content validator works fine. Haven't checked for the other validators yet.

PR #50, opened 2019-11-14
Better error messages regarding content_type

giovannibonetti:
Thanks for building this gem. We were working with error messages when we realized that `authorized_types` returns the MIME type, which is not very friendly to a non-technical user:

```ruby
authorized_types: 'application/pdf, image/jpeg, image/png'
```

Hence we've done a very small change to make it friendlier:
```ruby
authorized_types: 'PDF, JPEG, PNG'
```

That can be used in the error message like this:
```yml
content_type_invalid: "is not one of the valid types: %{authorized_types}"
```

### DIFF
--- a/lib/active_storage_validations/content_type_validator.rb
+++ b/lib/active_storage_validations/content_type_validator.rb
@@ -26,7 +26,9 @@ module ActiveStorageValidations
     end
 
     def types_to_human_format
-      types.join(', ')
+      types
+        .map { |type| type.to_s.split('/').last.upcase }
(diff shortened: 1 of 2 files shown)


igorkasyanchuk:
merged, thank you. soon, a new version will be released

giovannibonetti:
Thank you! 👍

igorkasyanchuk:
released, try it and see if all works

giovannibonetti:
It works! We have deployed it to production already! 👍

PR #54, opened 2020-01-15
Create fr.yml

dlepage:
Here a new locale to add to the list ! :)
Many thanks for this great gem.


igorkasyanchuk:
merged! thank, expect a new version later today

Issue #48, opened 2019-11-05
Rspec test matchers?


ImOnMars:
This isn't really an issue per se so feel free to close it ASAP, but I was just wondering if there's a way to test the validations in Rspec, similar to how Paperclip had `validate_attachment_content_type`?

igorkasyanchuk:
Don't know. But open for a PR's to merge them. If you can take a look how it's implemented in Paperclip and prepare PR I'll merge it

ImOnMars:
I will look into it but I'm brand new to ActiveStorage and have no experience writing matchers.

igorkasyanchuk:
use similar paperclip or carrierwave matchers as an example

ImOnMars:
Okay, well I'm not necessarily going to PR them as I'm unsure of whether these are good or extensive implementations, as well as the fact that I haven't tested them very much, but I seem to have implemented a couple of them:

Attachment Presence:
```ruby
# Validate attached file prescence matcher for ActiveStorage
RSpec::Matchers.define :validate_attachment_presence do |name|
  match do |record|
    record.send(name).purge # Delete attachment
    record.valid? # Rerun validations
    record.errors[:"#{name}"].present? # Check for validation error when attachment isn't present
  end
end
```

Attachment Content Type:
```ruby
# Validate attached file content type matcher for ActiveStorage
# content_types & excluded_content_types: Accepts an array of content_type strings
# not_content_types is optional:
#   if you leave it empty, it verifies that all common content types (excluding passed valid `content_types`) are invalid
#   if you pass in an array of content types, it will verify that those content types are invalid
#   if you pass false, it does not verify that other content types are invalid at all
#
# Note: 'common content types' are defined in AttachmentHelper
RSpec::Matchers.define :validate_attachment_content_type do |name, content_types, excluded_content_types = nil|
  not_content_types = case excluded_content_types
                        when false then false
                        when nil then AttachmentHelper.common_types - content_types
                        else excluded_content_types
                      end

  match do |record|
    file_path = "#{Rails.root}/spec/fixtures/files/empty_file"
    file_name = 'empty_file'

    content_types.each do |content_type|
      record.send(name).attach(
        io: File.open(file_path),
        filename: file_name,
        content_type: content_type
      )
      record.valid?

      return false if record.errors[:"#{name}"].present?
    end

    return true unless not_content_types

    not_content_types.each do |not_content_type|
      record.send(name).attach(
        io: File.open(file_path),
[...]

PR #56, opened 2020-01-24
Add test matchers for validators

StefSchenkelaars:
I worked on a first version of the test matchers as requested in https://github.com/igorkasyanchuk/active_storage_validations/issues/48. For now I've only implemented:
- `AttachedValidatorMatcher`
- `ContentTypeValidatorMatcher`
- `DimensionValidatorMatcher`
- `SizeValidatorMatcher`

and they should also be extended more to test all options of the validators (like message and validation context). But I hope it is a starting point and that we can add a complete set of matchers for all validators.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,11 +226,11 @@ describe User do
   it { is_expected.to validate_dimensions_of(:avatar).width_between(200..500) }
   it { is_expected.to validate_dimensions_of(:avatar).height_between(100..300) }
 
-  it { is_expected.to validate_dimensions_of(:avatar).less_than(50.kilobytes) }
-  it { is_expected.to validate_dimensions_of(:avatar).less_than_or_equal_to(50.kilobytes) }
-  it { is_expected.to validate_dimensions_of(:avatar).greater_than(1.kilobyte) }
-  it { is_expected.to validate_dimensions_of(:avatar).greater_than_or_equal_to(1.kilobyte) }
-  it { is_expected.to validate_dimensions_of(:avatar).between(100..500.kilobytes) }
+  it { is_expected.to validate_size_of(:avatar).less_than(50.kilobytes) }
+  it { is_expected.to validate_size_of(:avatar).less_than_or_equal_to(50.kilobytes) }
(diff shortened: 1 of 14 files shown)


igorkasyanchuk:
@StefSchenkelaars this is great news, as I see it's created for mini-test, which I think good start. Maybe you or someone else in the future could implement the same for RSpec.

StefSchenkelaars:
@igorkasyanchuk Well, I actually build it to work with `RSpec` since that what we use. I am not an expert in minitest but I think they should also work with minitest.

StefSchenkelaars:
I've tested the matchers with minitest and the shoulda context gem. Everything worked as expected and I've updated the readme accordingly.

igorkasyanchuk:
This is fantastic!

igorkasyanchuk:
let me know if you think it's ready for a review

StefSchenkelaars:
Worked a bit on making sure that the stubbing works with both Minitest and RSpec. I now think that works. Also made sure that de dimension validator matcher works with exact matches. 

As I said before, there is plenty of work to be done to make the matchers great but they help me quite a bit already in the current state. 

But it would be great if you can do a code review! 

And a bit unrelated is the rubocop. I ran it on the codebase and there where many failures. But I leave that up to another PR to fix (guess it's a good idea to add rubocop to travis)

igorkasyanchuk:
merged and released a new version. Thank you

connorshea:
This is great, thanks! 👍 

FWIW, pundit-matchers auto-includes itself if the matcher file is required, not sure if there are downsides to that but I figured I'd mention it: https://github.com/chrisalley/pundit-matchers/blob/00c89e7fa1f4dde6f2ad0c03a79db629a8f99e3e/lib/pundit/matchers.rb#L347-L351

StefSchenkelaars:
@connorshea It's probably a good idea to write a new issue for this and include your proposal. Since I think it will get lost since you commented on a closed PR. If I get bored sometimes, I might have a look at this, but you might fix it yourself too ;)

Issue #55, opened 2020-01-22
Storing blob when validation fails


anquinn:
Rails version: 6.0.2.1
Gem version: 0.8.6

Validation:
`validates :bill_kit, content_type: { in: ['application/pdf', 'application/msword', 'application/vnd.openxmlformats-officedocument.wordprocessingml.document'], message: 'is not a PDF or Word Document' }`

If you upload a file other than a word doc or a PDF the validation fails (as it should) but the blob is still stored in the db. The attachment is not created.

igorkasyanchuk:
@anquinn can you try to make a fix? or at least add a test that fails and reproduces your problem?

anquinn:
This is my first time using active_storage, still figuring it all out. Just wanted to report the bug.

Mate2xo:
I have the same problem on Rails 5.2. I initially made a custom `ContentTypeValidator` that had a `value.purge` statement to delete the blob and the  file from strorage. The problem was that, apparently, `.purge` relaunches all validations : this unexpected behavior made it difficult to populate the `errors` array properly, as the purged record being re-validated is in a different state upon 2nd validation.
So then I came to find your gem (which is awesome, thank you), but I notice the following behavior from ActiveStorage:
* On a new record initialization, assigning a file to a new instance (without saving it!) *immediately* uploads the file to storage (I believe this is fixed in Rails 6)
* When validation fails, the instance is not saved, but the blob is, and the file remains in storage. So in the end, an invalid file is still uploaded to, and kept in storage.

So for now, I'll manually instruct my controller to `.purge` on invalidation, as I don't know how to purge inside the validator without messing with this "double validation launch" behavior. Sorry I can't suggest you a fix for this.

I'll try to write a test for this later. In the meantime you should be able to reproduce this behaviour easily in the rails console :
* assign an invalid file to a new record, and you should see the file immediately uploaded.
* save the new record to see validation fail, but the blob and file remain strored

igorkasyanchuk:
@Mate2xo yes, this is a strange issue, and I don't have a fix.

So a possible solution would be is do create at least some snippet of code that could be used inside gem or app to cleanup attachments.  I like your idea of adding a spec for this and it would be a good start to fix this issue. Your contribution is more than welcome!

Mate2xo:
OK I'll try some things and share the results

igorkasyanchuk:
Can someone help with the validation of this PR? 
And maybe this "purge" should be added to other validators?

[...]

PR #64, opened 2020-02-25
Rails 5.2 attachement purge on content type invalidation

Mate2xo:
Hi,
Following the #55 issue, here are 2 commits to get started:
* The first one adds a failing test to check that an attached file with an invalid content type gets purged in rails 5.2
* The second one is a lead that I used in my project (I had to use `.purge` in a controller), but as mentioned in the issue, it apparently messes up the validation callbacks. In this case, this messes up another seemingly unrelated test. You can use it to see where it fails, but another solution seems to be necessary. So this 2nd commit is just an example.

I'm still a beginner, so I hope this isn't too 'dirty'. Also, I would love to know your solution to this problem :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ end
 To run tests in root folder of gem:
 
 * `BUNDLE_GEMFILE=gemfiles/rails_5_2.gemfile bundle exec rake test` to run for Rails 5.2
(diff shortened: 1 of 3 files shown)


igorkasyanchuk:
closing this for now, I think it's not a part of validator, and issue is fixed in newer rails versions as I understand

PR #65, opened 2020-03-04
Spelling error for authorized_types variable

mohanklein:
Readme says "authorized_type" instead of "authorized_types"


igorkasyanchuk:
thank you!

PR #68, opened 2020-04-04
Do not require attachments to be attached to use limit validator

rstammer:
The dummy app's `Project` class makes use of the `LimitValidator` by specifying the minimum and maximum count of attached documents:

```Ruby
class Project < ApplicationRecord
  …
  validates :documents, limit: { min: 1, max: 3 }
end
```

I noticed that one currently (without this PR) needs to add `attached: true` to employ the ` AttachedValidator` to surpass the first guard line within `LimitValidator`:

```Ruby
module ActiveStorageValidations
  class LimitValidator < ActiveModel::EachValidator # :nodoc:
    … 
    def validate_each(record, attribute, _)
      return true unless record.send(attribute).attached? # <--- Enforces the AttachedValidator
      …
    end
    …
  end
end
```
When I understood everything correctly (please correct me otherwise 🙈) then I think it's simply not necessary to do so and the `AttachedValidator` can left out here. Doing so and leaving it out makes the documented setup, that is this line:

```Ruby
validates :documents, limit: { min: 1, max: 3 }
```

work as expected. If it's not left out one would need to add `attached: true` to the above statement to make it work.

@igorkasyanchuk In case I misunderstood this and `attached: true` should be better added in such cases, just let me know!

As I expect that one does not necessarily need the `AttachedValidator` here, this PR drops the guard and fixes some test case situation, including a specific test edge case for the case of checking for the minimum value.


igorkasyanchuk:
@neumanrq sorry for a very long response just missed that

if you remove that line, can you validate that min: 0 + max: 3?

do you see any other impact with this change?

neumanrq:
Hello @igorkasyanchuk! Thanks for your feedback. By the time I proposed it, I did not see any further impact by this change. Unfortunately, as I have switched my job recently, I have no access to the repository where we used this patched version. Maybe @elalemanyo can check if there's time and if still in use?

elalemanyo:
@igorkasyanchuk We can validate that min: 0 + max: 3 and as @rstammer already said there is not impact 😊

UliKuch:
Hi @igorkasyanchuk, just wanted to ask if you plan on merging this 🙂

gr8bit:
I think this PR is quite important as it makes the limit validation much more intuitive. @rstammer @elalemanyo can one of you resolve the conflicts of this branch?

rstammer:
@gr8bit I will dive in and return with and push the updated branch!

gr8bit:
@rstammer HNY! 😃 I'd love to have this PR in the next release, would you have the time to sync and check the conflicts anytime soon?

rstammer:
Hi @gr8bit - sorry for dropping the ball here again. I will try to dive in tomorrow and fix the test file!

rstammer:
Hey @gr8bit - I've resolved the merge conflicts and fixed the tests. However, as the original idea for this arose two years ago, I'd appreciate you double-checking this to verify it's still meaningful. Thanks!! 🙌

test/active_storage_validations_test.rb
```diff
@@ -352,7 +369,10 @@ class ActiveStorageValidations::Test < ActiveSupport::TestCase
     e.proc_small_file.attach(dummy_file)
     e.attachment.attach(pdf_file)
     e.proc_attachment.attach(pdf_file)
+    e.documents.attach(pdf_file)
+    e.proc_documents.attach(pdf_file)
     e.dimension_exact.attach(image_150x150_file)
+    e.proc_documents.attach(pdf_file)
```
gr8bit, line 375: Duplicate to line 373?

gr8bit (changes requested):
Thank you @rstammer!
Except for this single change everything looks fine to me. I'd really like to have this fix in 1.1.0 (new minor version as it slightly changes limit behavior).

test/active_storage_validations_test.rb
rstammer: Good catch. Thanks! Fixed now ✌️

gr8bit (approved):
All fine, thanks a lot!

PR #73, opened 2020-05-15
I18n corrections for German language

mschnitzer:
This commit rephrases some of the error messages for the German translation.

As a native speaker, some of the sentences sounded a bit weird to me. I think it's better to rephrase them.


igorkasyanchuk:
merged, thank!

igorkasyanchuk:
active_storage_validations-0.8.9.gem released

PR #76, opened 2020-05-24
Gems versions bumped

reckerswartz:
Bumped bundler to 2.1.4
basic gem version update

fixed warnings like:- 

> Your Gemfile lists the gem coffee-rails (>= 0) more than once.
> You should probably keep only one of them.
> Remove any duplicate entries and specify the gem only once.
> While it's not a problem now, it could cause errors if you change the version of one of them later.


igorkasyanchuk:
merged, thanks

Issue #78, opened 2020-08-01
Error on attached matcher


alejandrodevs:
I have the following model:

```ruby
class Payment < ApplicationRecord
  validates :voucher, attached: true, if: :paid?
end
```

And the following test:

```ruby
context 'when is paid' do
  subject { build(:payment, status: :paid) }

  it { is_expected.to validate_attached_of(:voucher) }
end

context 'when is pending' do
  subject { build(:payment, status: :pending) }

  it { is_expected.to_not validate_attached_of(:voucher) }
end
```

it fails in the "when is paid" context. I was checking the code out and I think the error is here:
https://github.com/igorkasyanchuk/active_storage_validations/blob/master/lib/active_storage_validations/matchers/attached_validator_matcher.rb#L35

You are always creating a new instance of the subject class:
https://github.com/igorkasyanchuk/active_storage_validations/blob/master/lib/active_storage_validations/matchers/attached_validator_matcher.rb#L35

In order to support this it should be: `@subject = subject` instead of getting the class.

igorkasyanchuk:
Hi @alejandrodevs, can you make a PR with a fix?

alejandrodevs:
Sure, not problem. #79 please validate.

Mth0158:
Since I think this issue has been solved in #81 as mentioned in the issue thread, I am closing this ticket.
Let us know if not the case so we can reopen it.

PR #79, opened 2020-08-04
Fix attached matcher with a contextual subject

alejandrodevs:
Refer to: #78


alejandrodevs:
@igorkasyanchuk Could you please take a look?

PR #81, opened 2020-09-16
Ensure to detach attachment before validating attached

StefSchenkelaars:
It's a bit related to #78. In the case where you provide an instance which already has an attachment to the attached validator, you get a failure. So this example will fail while the attached validation is actually in place:

```ruby
subject { build :user, :with_avatar } # You get the point of this being factory bot that builds a user with an avatar attached
it { is_expected.to validate_attached_of :avatar } # This then fails
```

This PR fixes that case. And while I was on it I've also added additional checks to the matchers (I should have probably made a separate PR for this 🙈 ).

So hope you get the idea and let me know if there are any questions!

### DIFF
--- a/gemfiles/rails_5_2.gemfile.lock
+++ b/gemfiles/rails_5_2.gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
-    active_storage_validations (0.8.9)
+    active_storage_validations (0.9.0)
       rails (>= 5.2.0)
 
 GEM
--- a/gemfiles/rails_6_0.gemfile.lock
+++ b/gemfiles/rails_6_0.gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
-    active_storage_validations (0.8.9)
+    active_storage_validations (0.9.0)
       rails (>= 5.2.0)
 
 GEM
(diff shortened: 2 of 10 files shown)


igorkasyanchuk:
merged, I'll release a new version today, thanks

igorkasyanchuk:
oh, I didn't notice, but the build is failed: https://travis-ci.org/github/igorkasyanchuk/active_storage_validations
could you please check it and make a fix? @StefSchenkelaars

StefSchenkelaars:
@igorkasyanchuk Uhm, I now wanted to have a look at it but it's already merged. You fixed the failing spec yourself or did I miss something here? But thanks for merging 👍

igorkasyanchuk:
hello @StefSchenkelaars , actually I didn't have time to fix it. Could you please take a look into it?
I have a plan to release a new version in the near days I want to have all specs green. Thank you and sorry for asking, but I just can't do it on my own now.

igorkasyanchuk:
Error is here: https://travis-ci.org/github/igorkasyanchuk/active_storage_validations

StefSchenkelaars:
Ah, it looks like a rails 6.1 upgrade error. I don't think I'll be able to look at this today or tomorrow but afterwards the Christmas holiday starts ;) If anyone has any clues on what changed exactly and/or how to fix it, let me know!

igorkasyanchuk:
@StefSchenkelaars that error was before 6.1 merge, specs actually running with an older version of Rails. Sure, once you have time, I just want to release it sooner :)

PR #90, opened 2020-11-06
add es.yml translation

yshmarov:


igorkasyanchuk:
the new version released, thanks!

PR #96, opened 2020-12-20
Fix build

StefSchenkelaars:
As discussed in https://github.com/igorkasyanchuk/active_storage_validations/pull/81, I would fix the current build. 

The solution consisted of two steps:
- The addition of the Spanish translations where added with an incorrect key.
- My PR with the attached validator didn't properly unset the relation for unpersisted records.

I think it's a good idea to properly test the matchers on a real world codebase before putting out the release. I'll try running it against one of our apps.

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -1,4 +1,4 @@
-en:
+es:
   errors:
     messages:
       content_type_invalid: "tiene un tipo de contenido inválido"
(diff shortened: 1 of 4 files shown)


StefSchenkelaars:
@igorkasyanchuk pinging you just in case you missed it

igorkasyanchuk:
merged 👍 , thanks

PR #100, opened 2020-12-22
Fix deprecation warnings for Ruby 2.7

### LINKED ISSUE #98: Ruby 2.7.2 deprecation warning for Ruby 3
I'm pretty sure this is a problem in active_storage_validations, not with activemodel, though I might be wrong.

From running my specs on Ruby 2.7.2 with the `RUBYOPT='-W:deprecated'` environment variable set:

```
/Users/connorshea/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/bundler/gems/active_storage_validations-b84dfeaba49d/lib/active_storage_validations/content_type_validator.rb:17: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/connorshea/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/activemodel-6.1.0/lib/active_model/errors.rb:404: warning: The called method `add' is defined here
/Users/connorshea/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/bundler/gems/active_storage_validations-b84dfeaba49d/lib/active_storage_validations/size_validator.rb:28: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/connorshea/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/activemodel-6.1.0/lib/active_model/errors.rb:404: warning: The called method `add' is defined here
```

(The reason that environment variable is necessary is because 2.7.2 made the deprecation warnings hidden by default).

The gem is on a commit SHA because it's on my branch for #94, but nothing has changed on master that should really make a difference for this, as far as I can tell.

With Ruby 3.0 coming out in a few days, I think it's a good idea to make sure the gem doesn't have any deprecation warnings like this. 

See this article for more information on the kwargs changes: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

connorshea:
Fixes #98

### DIFF
--- a/lib/active_storage_validations/content_type_validator.rb
+++ b/lib/active_storage_validations/content_type_validator.rb
@@ -14,7 +14,7 @@ module ActiveStorageValidations
         next if is_valid?(file)
 
         errors_options[:content_type] = content_type(file)
-        record.errors.add(attribute, :content_type_invalid, errors_options)
+        record.errors.add(attribute, :content_type_invalid, **errors_options)
         break
       end
     end
--- a/lib/active_storage_validations/dimension_validator.rb
+++ b/lib/active_storage_validations/dimension_validator.rb
@@ -123,7 +123,7 @@ module ActiveStorageValidations
       key = options[:message].presence || type
       return if record.errors.added?(attribute, key)
       record.errors.add(attribute, key, *attrs)
-    end    
+    end
 
   end
 end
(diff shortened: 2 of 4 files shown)


igorkasyanchuk:
new version releases

connorshea:
Thank you! I've updated and deployed and everything seems to work correctly :) 
Thanks!

PR #102, opened 2021-01-11
Rails 6 1

StefSchenkelaars:
Hi @igorkasyanchuk

I got some errors when trying to use this gem with rails 6.1. So I wanted to fix those tests. But in order to do that, the migrations are kinda tricky. So I've now changed the setup to use the [combustion gem](https://github.com/pat/combustion) so we can easily change the schema.

Have a look at this already and then I'll continue on fixing rails 6.1


igorkasyanchuk:
Hi @StefSchenkelaars 
this looks awesome

so you think it soon could be merged into master?

StefSchenkelaars:
Yea I'm fine with merging it. However travis is not helping... [it's not running my code](https://travis-ci.org/github/igorkasyanchuk/active_storage_validations/builds/753970165). So if that's all ok, you can merge.

Maybe good to know that I removed the dummy app completely and thus you cannot boot the rails app.

igorkasyanchuk:
merged! thanks!

PR #117, opened 2021-05-10
Add with_message for dimension_validator_matcher

jayshepherd:
The dimension validator matchers fail when a custom message is supplied.  This PR adds a `with_message` option to the matcher which will take precedence over the default message.


igorkasyanchuk:
thanks, I'll release a new version soon

Issue #86, opened 2020-09-29
Attached message


SirNerdBear:
When doing this:
`validates :stored_image, attached: true`

The error message is "Can't be blank"

Normally with rails I can do:
`validates :stored_image, attached: {message: "Must provide an image"}`

But that is not an option here. I was hoping one of the translation keys was for a missing image but doesn't appear that's the case. Unless I'm missing something?

PR #121, opened 2021-05-29
Enable to set message on error of attached validator

ohbarye:
This pull request tries to fix https://github.com/igorkasyanchuk/active_storage_validations/issues/86.


igorkasyanchuk:
thanks, I'll release an updated version of the gem soon

Issue #122, opened 2021-06-24
Attachments has limited mime types variations


randsina:
Rails: 6.1.3.2
ActiveStorageValidations: 0.9.4

The model:
```ruby
class QuestionResult < ApplicationRecord
  has_one_attached :document

  validates :document, content_type: %i[doc docx xls xlsx ppt pptx pdf pages numbers]
end
```
the tests
```ruby
require 'spec_helper'

describe QuestionResult do
  it do
    is_expected.to validate_content_type_of(:document).allowing(%w[
      application/msword
      application/vnd.openxmlformats-officedocument.wordprocessingml.document
      application/vnd.ms-excel
      application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
      application/vnd.ms-powerpoint
      application/vnd.openxmlformats-officedocument.presentationml.presentation
      application/pdf
      application/vnd.apple.pages
      application/vnd.apple.numbers
    ])
  end
end
```
got `content_type must be one of Regxep, supported mime types (e.g. :png, 'jpg'), or mime type String ('image/jpeg')`

It's possible to use `String` to define validations and have green tests but it's not so convenient.
```ruby
validates :document, content_type: %w[
  application/msword
  application/vnd.openxmlformats-officedocument.wordprocessingml.document
  application/vnd.ms-excel
  application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
  application/vnd.ms-powerpoint
  application/vnd.openxmlformats-officedocument.presentationml.presentation
  application/pdf
  application/vnd.apple.pages
  application/vnd.apple.numbers
]
```

The current gem uses `Mime` to find the correct content_type. In the same time, `rails` is dependency of it and `activestorage` uses `marcel` gem to define content_type.

`Marcel` has own defined types, kind of 
https://github.com/rails/marcel/blob/d970fdafdf6ed5456fbe0d537108bcee33adbb10/lib/marcel/tables.rb#L238-L240
```ruby
    'doc' => 'application/msword',
    'docm' => 'application/vnd.ms-word.document.macroenabled.12',
    'docx' => 'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
```
```ruby
Marcel::MimeType.for extension: 'numbers'
# => "application/vnd.apple.numbers"
```
Does it make sense to use `Marcel` to check the right _content_type_?
What do think about it?

igorkasyanchuk:
[...]

PR #123, opened 2021-06-25
Use Marcel to detect the correct content_type

randsina:
Related to the issue #122.
`ActiveStorage` uses `Marcel` gem to detect `content_type` of an attachment.
Idea is to use the same gem for validations.

https://github.com/rails/rails/blob/9053b8f77ffdb28f2494d773cec6fa1a84eceabe/activestorage/app/models/active_storage/blob/identifiable.rb#L22


lib/active_storage_validations/matchers/content_type_validator_matcher.rb
```diff
@@ -57,7 +57,7 @@ def allowed_types
       end
 
       def rejected_types
-        @rejected_types || (Mime::LOOKUP.keys - allowed_types)
+        @rejected_types || (Marcel::TYPES.keys - allowed_types)
```
randsina, line 60:
```ruby
Mime::LOOKUP.keys.count
# => 45
```
```ruby
Marcel::TYPES.keys.count
# => 892
```
It influences the speed of tests

igorkasyanchuk: how big is the impact?

randsina: Doesn't look like a big impact for
```sh
$ BUNDLE_GEMFILE=gemfiles/rails_6_1.gemfile bundle exec rake test
```
before:
```
Finished in 5.077188s, 12.8024 runs/s, 23.0443 assertions/s.
65 runs, 117 assertions, 0 failures, 0 errors, 0 skips
```
after
```
Finished in 5.293181s, 13.0356 runs/s, 22.6707 assertions/s.
69 runs, 120 assertions, 0 failures, 0 errors, 0 skips
```

igorkasyanchuk:
merged, thanks 👍 
I'll release a new version soon

lib/active_storage_validations/matchers/content_type_validator_matcher.rb
IvanShamatov: https://github.com/igorkasyanchuk/active_storage_validations/issues/127

IvanShamatov (commented):
it definitely affected my suite

PR #126, opened 2021-06-25
Add latest Rails support

randsina:
Rails 7 will be released soon.
Add support of it to check the compatibility of the gem with the latest version of Rails.


igorkasyanchuk:
new version was released

PR #128, opened 2021-07-07
Replace deprecated .create_after_upload!

randsina:
1. Use the same syntax and condition for `Rails.version`
2. Replace deprecated `Blob.create_after_upload!` with `Blob.create_and_upload!`


igorkasyanchuk:
merged 👍

PR #129, opened 2021-07-13
Only require necessary Rails subcomponents 

### LINKED ISSUE #106: Rails dependency
With the dependency on Rails, and not its sub-components, it's not possible for an app that uses this gem to cut out parts of Rails to save memory. Can this be changed? Glad to come up with a PR if you can tell me what they should be (ActiveModel?).

vietqhoang:
## Problem

The entire [Ruby on Rails](https://github.com/rails/rails) framework is required when utilizing `active_storage_validations`. Loading the entire RoR framework is unnecessary when the `active_storage_validations` project itself only uses a few Rails subcomponents. In addition, any project that is resource optimized to load only a subset of RoR subcomponents and depends on `active_storage_validation` is forced to include the entire RoR framework.

## Solution

Replace the dependency of `rails` in the gemspec to only include the necessary subcomponents.

Using [Rails' gemspec](https://github.com/rails/rails/blob/main/rails.gemspec#L31-L41) as reference, the following were identified as necessary by scanning the project codebase:

* ActiveModel
* ActiveStorage
* ActiveSupport

The `rail` dependency in the gemspec is replaced with the three subcomponents and the test Gemfiles were updated to reflect the new dependencies environment.

All tests passed except the Rails 5.2 environment. The following message resulted: `File does not exist: active_job/railtie`.

Adding `activejob` to the dependency list removed the roadblock of the Rails 5.2 environment tests to execute and subsequently passing.

---

Closes #106


igorkasyanchuk:
Thanks! 👍

PR #133, opened 2021-09-10
README update - better clarify how to define between size

equivalent:


gr8bit (approved):
Additions are correct. That's how :between can be used.

PR #134, opened 2021-09-28
Add min_size and max_size to :file_size_out_of_range error message

kemenaran:
Currently, the error message for an invalid file size can't include the expected min or max size of the file.

The error message only states:

>  "Attachment size %{file_size} is not between required range"

---

This PR allows using the interpolated variables `%{min_size}` and `%{max_size}` in the localized error message string.

Thus the error message can be customized by projects using this gem to look like this:

>  "Attachment size %{file_size} is larger than the required size %{max_size}"

## Future work

For now, the error message doesn't include the size information by default. The extra information must be added by projects using the gem.

In a future PR, I'd like to introduce new localized strings for "too large" and "too small" errors, which would allow to have the default error messages mention the string by default. (This is more work because of retro-compatibility concerns though).

### DIFF
--- a/lib/active_storage_validations/size_validator.rb
+++ b/lib/active_storage_validations/size_validator.rb
@@ -25,6 +25,9 @@ module ActiveStorageValidations
         next if content_size_valid?(file.blob.byte_size)
 
         errors_options[:file_size] = number_to_human_size(file.blob.byte_size)
+        errors_options[:min_size] = number_to_human_size(min_size)
+        errors_options[:max_size] = number_to_human_size(max_size)
+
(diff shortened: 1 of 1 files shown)


kemenaran:
@igorkasyanchuk would that be interesting for the project, and is there something I could do to move this PR forward?

igorkasyanchuk:
merged, sorry, missed it, today will release a newer version

igorkasyanchuk:
0.9.6 released

kemenaran:
Hey, that's great, thanks a lot!

PR #135, opened 2021-10-06
#105 Allow Procs for dynamic value usage

### LINKED ISSUE #105: Allow Procs as values for validator restrictions
Hi there!

Sometimes it is necessary to have dynamic values for validations. I have seen in some other validator gems also procs can be passed instead of integers. These are then called in the validator methods. Then it is possible for example to use ENV variables to set the dimensions or sizes.

That would be another super addition to make this extension even more useful.

codegeek319:
This PR allows Procs for the value determinations at runtime. Also the Test and the Documentation is updated according to the latest code.
closes #105


lib/active_storage_validations/content_type_validator.rb
```diff
@@ -31,7 +35,7 @@ def types
       end
     end
 
-    def types_to_human_format
+    def types_to_human_format(types)
```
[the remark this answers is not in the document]
gr8bit: @igorkasyanchuk sorry for the late answer and tackling of this issue!
For the dynamic proc evaluation, we need to check the options for procs on every validation, call them, then continue the validations with the new options. As the options are used in the `types` method to determine the authorized types, we now have to pass the new options to it and the resulting authorized types can be different on each validation - that's why we then pass the types on to the methods which need them.

I renamed the `types` method to `authorized_types` to avoid confusion with variables and parameters. Looks clearer now, I agree. :)

gr8bit:
@igorkasyanchuk would you mind to take another look? Thanks! :)

igorkasyanchuk:
Hello @gr8bit ,

Sorry for taking so long time, I think I've some important comments now.
I see in the code that often you use something like `range = range.call(record) if range.is_a?(Proc)` and you checking in many places objects if it's proc or now. Because of it you changed lot's of code, and added some complexity to existing method, for example by adding new parameters.

I've the following idea, it probably can take not long time to implement it, at least I want you to think if it could work better.
So the idea is the following - validation class receives some "options", which are consist of hash and potentially hashes inside.

We can just convert all hashes to values before. So no need to call proc somewhere inside validator. Everything will keep working as before.

I did some quick prototype, you can use it as a reference and improve. Ready to discuss it with you, If needed in skуpe "igorkasyanchuk".

image

igorkasyanchuk:
@codegeek319 sorry mentioning you too. Just in case you need:

```ruby
    def recur_call_hash(record, h)
      hash = h.dup
      return hash unless hash.is_a?(Hash)

      puts hash
      hash.each do |(k,v)|
        if v.is_a?(Proc)
          hash[k] = v.call(record)
        else
          hash[k] = v
        end
        if hash[k].is_a?(Hash)
          hash[k] = recur_call_hash(record, hash[k])
        end
      end
      hash
    end
```

igorkasyanchuk:
and last suggestion, to make in Readme example more close to real life if possible. For example

```ruby
validates :proc_files, limit: { max: -> (record) { record.admin? ? 100 : 10 } }
```

gr8bit:
Thank you for the great feedback!! @codegeek319 has been implementing this and we'll discuss it on Tuesday, we work together irl. :)

igorkasyanchuk:
@gr8bit @codegeek319 do you have plans to fix this PR?

gr8bit:
Sure thing! Sorry, we have been quite busy ☺️

gr8bit:
@igorkasyanchuk finally I found the time for the refactoring you suggested. Diff shrunk significantly. Would you mind taking another look? :)

igorkasyanchuk:
thank you for your PR, from my side I think it looks good

But I want to check with other developers who did the most contributions to the gem

@StefSchenkelaars , @phlegx , @tleneveu , @reckerswartz what do you think about this PR?

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -24,11 +26,12 @@ def validate_each(record, attribute, _value)
         changes = record.attachment_changes[attribute.to_s]
         return true if changes.blank?
 
+        options = unfold_procs(record, self.options, AVAILABLE_CHECKS)
```
StefSchenkelaars, line 29: I don't like changing the original options variable. The `options` variable is made available in the `ActiveModel::Validator` and I would really keep it that way.

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -4,6 +4,8 @@
 
 module ActiveStorageValidations
   class AspectRatioValidator < ActiveModel::EachValidator # :nodoc
+    include OptionProcUnfolding
```
StefSchenkelaars, line 7: Can we reuse the code and logic from active model? Since we depend on it anyway, why not use that a bit more. For example the https://github.com/rails/rails/blob/d695972025146713fae9a089dcaf2239ede354d4/activemodel/lib/active_model/validations/resolve_value.rb#L6 method might be useful.

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -24,11 +26,12 @@ def validate_each(record, attribute, _value)
         changes = record.attachment_changes[attribute.to_s]
         return true if changes.blank?
 
+        options = unfold_procs(record, self.options, AVAILABLE_CHECKS)
         files = Array.wrap(changes.is_a?(ActiveStorage::Attached::Changes::CreateMany) ? changes.attachables : changes.attachable)
 
         files.each do |file|
           metadata = Metadata.new(file).metadata
-          next if is_valid?(record, attribute, metadata)
+          next if is_valid?(record, attribute, metadata, options)
```
StefSchenkelaars, line 34: This `options` is an instance variable (or at least it should be). So passing this should be unnecessary. If the unfolding of the options has to be done, do that in the location where you need it (in the private method).

lib/active_storage_validations/content_type_validator.rb
```diff
@@ -2,27 +2,33 @@
...
         break
       end
     end
 
-    def types
-      return @types if defined? @types
-
-      @types = (Array.wrap(options[:with]) + Array.wrap(options[:in])).compact.map do |type|
+    def authorized_types(options)
+      (Array.wrap(options[:with]) + Array.wrap(options[:in])).compact.map do |type|
```
StefSchenkelaars, line 31: This is an instance method but is actually used as a helper. I would really try to make clear that these authorized types are dependent on the `record` (which they are implicitly since the options depend on the record). I would therefore rewrite this method to something like this:

```ruby
def authorized_types(record)
  with = resolve_value(record, options[:with])
  in = resolve_value(record, options[:in])
  
  (Array.wrap(with) + Array.wrap(in)).compact.map do |type|

```

This makes it clear that the authorized types depend on the `record` and you don't change the original `options`.

StefSchenkelaars (commented):
I really like the idea! Makes it more similar to Active Model so that's good.
Code wise I think it should also be a bit more like Active Model. Especially the local version of the `options` variable seems dangerous to me since it gives quite some ambiguity.
So try to ensure the instance methods are actually instance methods and are not used as some helper method which still accesses instance variables. 
If you have any questions about my comments or need a bit of help implementing, let me know!

lib/active_storage_validations/aspect_ratio_validator.rb
gr8bit: That method unfortunately isn't yet part of any rails stable version. We're downwards compatible to 5.2 at the moment, right? We might copy it though...

lib/active_storage_validations/aspect_ratio_validator.rb
gr8bit: I hope I understand you correctly but the original `options` variable will not be changed, it will be masked by a local variable called `options`. To avoid this confusion, I'll rename the local variable with `flat_options`.

lib/active_storage_validations/aspect_ratio_validator.rb
gr8bit: No it is not the instance variable, it's the local variable `option` which was masked above. Which I will rename to `flat_options` to make it clearer.

lib/active_storage_validations/content_type_validator.rb
gr8bit: Actually, the authorized types depend on the options (=flat_options now) which have to be passed in. We're using this same pattern in every other validator:
```ruby
flat_options = unfold_procs(record, self.options, AVAILABLE_CHECKS)
```
The `AVAILABLE_CHECKS` constant specifies which keys from the options should be resolved (which, in this case, are `%i[with in]`).

gr8bit:
@StefSchenkelaars I commented on and coded towards your suggestions. :) What do you think now?

StefSchenkelaars:
I still don't really like the passing of these options all the time but I'm ok if it works.

lib/active_storage_validations/content_type_validator.rb
StefSchenkelaars: I don't agree, the authorized types in the end depends on the record (and the options instance variable but that doesn't have to be passed). 

You can generate these flat options if you have the options and the record.

gr8bit:
> I still don't really like the passing of these options all the time but I'm ok if it works.

I'm totally open to suggestions! Please check the validators and find a better/easier way - keep in mind though the options have to be processed on every validation.

lib/active_storage_validations/content_type_validator.rb
gr8bit: I see, I just moved the proc unfolding call into the authorized_types and pass the record inside that instead of the unfolded options.

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -54,45 +53,45 @@ def validate_each(record, attribute, _value)
     private
 
 
-    def is_valid?(record, attribute, metadata)
+    def is_valid?(record, attribute, metadata, flat_options)
```
StefSchenkelaars, line 56: This additional depencency is not needed for this method. You can calculate the flat_options in this method since you have the record already.

lib/active_storage_validations/dimension_validator.rb
```diff
@@ -51,64 +57,65 @@ def validate_each(record, attribute, _value)
...
-          next if is_valid?(record, attribute, metadata)
+          next if is_valid?(record, attribute, metadata, flat_options)
           break
         end
       end
     end
 
 
-    def is_valid?(record, attribute, file_metadata)
+    def is_valid?(record, attribute, file_metadata, flat_options)
```
StefSchenkelaars, line 73: Same here, you already have the record. Less arguments = better :)

lib/active_storage_validations/dimension_validator.rb
```diff
@@ -120,10 +127,10 @@ def is_valid?(record, attribute, file_metadata)
       true # valid file
     end
 
-    def add_error(record, attribute, type, **attrs)
-      key = options[:message].presence || type
-      return if record.errors.added?(attribute, key)
-      record.errors.add(attribute, key, **attrs)
+    def add_error(record, attribute, default_message, **attrs)
+      message = options[:message].presence || default_message
```
StefSchenkelaars, line 131: Maybe an interesting possibility is to allow a proc for the message as well. But lets do that in a separate PR if people would like that idea.

StefSchenkelaars (commented):
Looks good, some remarks in the same line.

lib/active_storage_validations/content_type_validator.rb
StefSchenkelaars: I find this a lot clearer. Hopefully, you don't mind me pushing on this but I find it important that the lowest level of dependencies (arguments) for each function are used. This way it is harder to screw up, for example by passing the `flat_options` that were generated with a different record.

lib/active_storage_validations/dimension_validator.rb
gr8bit: Agreed and changed :)

lib/active_storage_validations/aspect_ratio_validator.rb
gr8bit: True, I changed it.

gr8bit:
I think this PR could be merged now.

StefSchenkelaars:
@igorkasyanchuk Agree with @gr8bit that it's good for a merge. I think you are the only one who can do that.

igorkasyanchuk:
Great) will do it later today
Thanks all for tremendous amount of work

igorkasyanchuk:
active_storage_validations-1.0.0.gem released

igorkasyanchuk:
looks like one regression https://github.com/igorkasyanchuk/active_storage_validations/issues/166 if someone can take a look on this

PR #137, opened 2021-10-19
Reference Marcel::EXTENSIONS

jrmhaig:
Active Storage 6.1.3.2 upgraded `marcel` to version 1.0, which removed the dependency on `mimemagic`. `MimeMagic::EXTENSIONS` and `MimeMagic.by_extension` have been replaced by `Marcel::EXTENSIONS` and `Marcel::Magic.by_extension` respectively. The README is updated accordingly, including notes for use with older versions.

PR #138, opened 2021-12-07
References `Marcel::TYPES` only for Rails < 6.1

luiseugenio:
Checking the reference to `Marcel::TYPES` only for Rails < 6.1.
Fix #137


igorkasyanchuk:
@luiseugenio thanks for the PR
what do you think about adding gem marcel to gemspec? to make it mandatory?

luiseugenio:
> @luiseugenio thanks for the PR what do you think about adding gem marcel to gemspec? to make it mandatory?

@igorkasyanchuk I think it makes sense. I will add and update the PR.

luiseugenio:
Done @igorkasyanchuk

PR #148, opened 2022-03-28
Fix a file extension issue in Metadata#read_image

stephensolis:
It turns out that `Vips::get_suffixes` only includes lowercase extensions. So, if you have a file with an extension eg. `.JPG`, the code would fall back to MiniMagick which is not desired.


igorkasyanchuk:
merged, thanks, I'll release a new version soon

Issue #146, opened 2022-03-10
Only one image validation error showing at a time


AlecRust:
Given an image attachment that is validated to minimum 320x320px:

```ruby
validates :headshot_image, content_type: %i[png jpg jpeg], dimension: { width: { min: 320 }, height: { min: 320 } }
```

When I upload a 20x20px image:

![Screen Shot 2022-03-10 at 15 44 03](https://user-images.githubusercontent.com/247634/157698958-b2639e97-aae2-484f-94ad-f817c7ac253e.png)

Then only the _width_ validation appears:

![image](https://user-images.githubusercontent.com/247634/157699352-c6241937-e9d9-442a-8019-8c470330a2aa.png)

If the width of the image is then increased (that's all the validation error says to fix) then the _height_ validation appears:

![image](https://user-images.githubusercontent.com/247634/157699575-7ab1d9fc-21ed-409b-91df-34dd659b7c55.png)

Is it possible for this gem to display all the validation errors at once to prevent this UX issue?

igorkasyanchuk:
Hi, I think it's possible, but you should make a PR with a fix or wait for someone who will do.

I think you should try to do it, instructions in the readme.

AlecRust:
This seems fixed now, thanks @Animesh-Ghosh!

PR #152, opened 2022-04-02
Display All Validation Errors

Animesh-Ghosh:
Display width and height validations errors together if both are invalid.

This is related to #146.


igorkasyanchuk:
@Animesh-Ghosh can you add a simple test for this PR?

Animesh-Ghosh:
We can also add [`simplecov`](https://github.com/simplecov-ruby/simplecov) to check the code coverage.
I had added it as a development dependency while writing the test and it seems the code coverage on `master` is **88.85%** and after this test was added it increase to **89.32%**.

Let me know if this is something that might be needed.

igorkasyanchuk:
@Animesh-Ghosh  simplecov would be nice to add. Please add it as a separate PR

PR #158, opened 2022-04-19
Add simplecov as a Development dependency

Animesh-Ghosh:
Added [`simplecov`](https://github.com/simplecov-ruby/simplecov) as per [this](https://github.com/igorkasyanchuk/active_storage_validations/pull/152#issuecomment-1100852295) comment.

### DIFF
--- a/active_storage_validations.gemspec
+++ b/active_storage_validations.gemspec
@@ -26,6 +26,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rubocop'
   s.add_development_dependency 'sqlite3'
   s.add_development_dependency 'marcel'
-  s.add_development_dependency 'simplecov', '~> 0.21.2'
+  s.add_development_dependency 'simplecov'
   s.add_development_dependency 'globalid'
 end
--- a/active_storage_validations.gemspec
+++ b/active_storage_validations.gemspec
@@ -26,4 +26,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rubocop'
   s.add_development_dependency 'sqlite3'
   s.add_development_dependency 'marcel'
+  s.add_development_dependency 'simplecov', '~> 0.21.2'
 end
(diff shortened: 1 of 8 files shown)


StefSchenkelaars:
@Animesh-Ghosh I don't think adding a `.ruby-version` makes sense since there is a test matrix for multiple ruby versions: https://github.com/igorkasyanchuk/active_storage_validations/blob/master/.github/workflows/minitest.yml#L9-L13

It might also be interesting to combine this addition with some CI pipeline to communicate how the new PR impacts the code coverage.

gemfiles/rails_5_2.gemfile.lock
```diff
@@ -51,6 +51,7 @@ GEM
       thor (>= 0.14.6)
     concurrent-ruby (1.1.7)
     crass (1.0.6)
+    docile (1.4.0)
```
StefSchenkelaars, line 54: Where did this one came from?

Animesh-Ghosh: Seems like it's one of `simplecov`'s dependencies.
https://github.com/simplecov-ruby/simplecov/blob/efdb08db63b35577b3b0d79db893ead1d848d8dd/simplecov.gemspec#L37

Animesh-Ghosh:
@igorkasyanchuk conflicts fixed.

Apologies for the delay!

active_storage_validations.gemspec
```diff
@@ -26,5 +26,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rubocop'
   s.add_development_dependency 'sqlite3'
   s.add_development_dependency 'marcel'
+  s.add_development_dependency 'simplecov', '~> 0.21.2'
```
gr8bit, line 29: Why the version specification here? Won't it work with 0.22.0? I'd rather keep the version open here as the specific versions will be in the .lock-files anyway.

gr8bit: @Animesh-Ghosh bird is the word!

Animesh-Ghosh: @gr8bit removed `simplecov` version specification from `active_storage_validations.gemspec`.

gr8bit:
Thanks a lot @Animesh-Ghosh! 🥳

PR #161, opened 2022-06-28
Aspect Ratio Validator error message should present aspect ratios as W:H, not WxH.

### LINKED ISSUE #160: Aspect Ratio generally includes a ':', not a 'x'
Aspect ratios are generally given as 16:9 or 4:3, not 16x9, at least in my experience (maybe this differs between language contexts). This results in the error messages this gem produces being a little confusing.

yarmiganosca:
Fixes #160 

I also did a very small amount of refactoring, by removing and unnecessary double-nested `else` and unnecessary escape characters from the aspect ratio regex.


yarmiganosca:
Tagging @corrieleech so she gets notified when this gets merged.

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -73,18 +73,18 @@ def is_valid?(record, attribute, metadata)
         return true if metadata[:width] > metadata[:height]
         add_error(record, attribute, :aspect_ratio_not_landscape)
 
-      else
-        if options[:with] =~ /is\_(\d*)\_(\d*)/
-          x = $1.to_i
-          y = $2.to_i
+      when /is_(\d*)_(\d*)/
```
gr8bit, line 76: These should probably be `(\d+)` as `is__1` or `is_5_` would be currently allowed.

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -73,18 +73,18 @@ def is_valid?(record, attribute, metadata)
...
-      else
-        if options[:with] =~ /is\_(\d*)\_(\d*)/
-          x = $1.to_i
-          y = $2.to_i
+      when /is_(\d*)_(\d*)/
+        x = $1.to_i
+        y = $2.to_i
 
-          return true if (x.to_f / y).round(PRECISION) == (metadata[:width].to_f / metadata[:height]).round(PRECISION)
+        return true if (x.to_f / y).round(PRECISION) == (metadata[:width].to_f / metadata[:height]).round(PRECISION)
```
gr8bit, line 80: If `y` will be supplied as 0 this line will cause a devision by zero. What do you think about adding that as condition?
```ruby
return true if !y.zero? && (x.to_f / y).round(PRECISION) == (metadata[:width].to_f / metadata[:height]).round(PRECISION)
```
The code will then default into the invalid aspect ratio error.

lib/active_storage_validations/aspect_ratio_validator.rb
gr8bit:
```ruby
when /\Ais_(\d*[1-9]\d*)_(\d*[1-9]\d*)\z/
```
would also ensure the string is exactly `is_X_Y` with X and Y being > 0.

gr8bit:
@yarmiganosca I commented on two things, what do you think about them? If you could fix the conflicts as well I'd totally second the PR. 👍

igorkasyanchuk:
@yarmiganosca PR looks good, and I also think that suggestions from @gr8bit make sense (since I know you just moved my personal code a little :))

if you can quickly apply suggestions I'll merge your PR

igorkasyanchuk:
oh, and also - you need to resolve conflicts :(

gr8bit:
@yarmiganosca are you still available for these fixes? Please let me know, I could finish this PR if you aren't. :)

yarmiganosca:
Sorry, I had completely forgotten I submitted this PR, so I thought I was just receiving messages from some repo I had commented on, not actually submitted the entire, ya know, PR. My apologies! I'll take a look later today.

gr8bit:
> Sorry, I had completely forgotten I submitted this PR, so I thought I was just receiving messages from some repo I had commented on, not actually submitted the entire, ya know, PR. My apologies! I'll take a look later today.

Sorry it took us so long! I'm glad we'll be able to merge these change soon. :)

yarmiganosca:
@gr8bit I think for misconfigurations of the validator itself (and I think what you're referring to would be a misconfiguration of the validator itself), we should try to detect those when the validator is declared on the model, not when validating individual records. I see we already have a check_validity! method, and it's my understanding that Rails will call that itself when a validation is declared on a model (at least I hope that's the case, because afaict check_validity! isn't actually called anywhere in this gem). If that understand is correct, then all these concerns you have are quite valid, but I think they should be addressed there instead of when we're validating individual records. What do you think?

gr8bit:
@yarmiganosca A very good point! I absolutely favor the approach of validating the input on declaration instead of during runtime.

Mth0158:
@yarmiganosca can you make the changes that @gr8bit required so we can merge this PR ? :)

Mth0158:
@igorkasyanchuk @gr8bit FYI

lib/active_storage_validations/aspect_ratio_validator.rb
Mth0158: I used `/is_([1-9]\d*)_([1-9]\d*)/` which is, I think, a bit better since it will not accept ratios like `is_01_2`

lib/active_storage_validations/aspect_ratio_validator.rb
Mth0158: Done :) even if this case should not occur with the new regex

Mth0158:
@yarmiganosca & @corrieleech FYI this will be released in the coming days / weeks

Issue #130, opened 2021-07-22
How i Can validate file by mime type


SynthesisOne:
Hello, tell me how you can check the file not only by extension but also by content, because at the moment, if, for example, change the file extension json to jpeg, then the validation does not work.
Thanks

igorkasyanchuk:
if you can create a PR with a fix

Mth0158:
Hi @Halvanhelv,
Since this issue has been solved by #168 by implementing a new validator `processable` ([check it here](https://github.com/igorkasyanchuk/active_storage_validations/blob/master/lib/active_storage_validations/processable_image_validator.rb)), I am closing this issue.
Let us know if you have any question

Issue #91, opened 2020-11-18
ActiveStorage accepts broken images and throws exceptions when creating variants


Signum:
**TL;DR** There is no way to detect broken image files. So exceptions are thrown later when ActiveStorage tries to render _variants_.

I originally submitted this issue to https://github.com/rails/rails/issues/40493 but was directed here.

### Steps to reproduce

Create a controller action that receives uploaded image files. A part from my Rails application:

```ruby
def upload_receive
  params[:file].each do |img|
    new_screenshot = @package.screenshots.new(image: img)
    
    if new_screenshot.valid?  
      new_screenshot.save
    end
  end
end
```

My model:

```ruby
class Screenshot < ApplicationRecord
  has_one_attached :image
  validates :image, attached: true, content_type: [:png]

  def medium_image
    if self.image.attached?
      self.image.variant(
        resize_to_limit: [670, 600]
      ).processed
    end
  end
end
```

(The validation uses the 'active_storage_validations' gem but the problem is most certainly not coming from there.)

This works for intact PNG files. A deliberately broken PNG file will be accepted and saved though. My test file is 5 KB of zero bytes that I named "broken.png" and uploaded to Rails.

### Expected behavior

One of…

1. ActiveStorage detect broken image files by trying to open them with Minimagick/Imagemagick as a validation.
1. ActiveStorage offers a common validation to detect broken images. (I tried to write my own custom validator to run the file through Minimagick but it I could not find a way to access the blob's data in the model before a ".save".)
1. Creating variants does not just throw an exception.

### Actual behavior

The file gets saved by ActiveStorage with content_type='image/png'. I suspect that 'marcel' and 'mimemagick' (which I believe are used for the content type detection) can't figure out what this file is and just look at the '.png' suffix.

Once a template tries to render a variant using the Screenshot.medium_image method it will call Minimagick and fail:

```
convert /tmp/ActiveStorage-21670-… failed with error:
convert-im6.q16: improper image header `/tmp/ActiveStorage-21670-…' @ error/png.c/ReadPNGImage/4092.
convert-im6.q16: no images defined `/tmp/image_processing….png' @ error/convert.c/ConvertImageCommand/3258.
```

### System configuration

**Rails version**: 6.0.3.4

**Ruby version**: 2.6.5p114

samiggapps:
I did find this answer on SO which helped fix my issue 
https://stackoverflow.com/a/60084096/7913423

Mth0158:
[...]

Issue #62, opened 2020-02-16
Security when using direct uploads?


janko:
When using direct uploads, ActiveStorage will delay identifying the file into a background job, spawned only after the file has been attached. Currently, the initial blob's `content_type` is  determined solely based on the extension of the uploaded file, and I believe this value will be the one that's used by `active_storage_vaidations`. In this case, if my app is accepting only images, a malicious user could upload a non-image file by giving it an image extension (e.g. jpg, png).

igorkasyanchuk:
I don't know what we can do to prevent this issue.  Maybe you can try to analyze the file later on the server and check if it's an image and if not - delete the file.

Gem doesn't support validation on the frontend, and of course, if you want to add it - all community will appreciate it.

vitobotta:
Hi @igorkasyanchuk So does this mean we cannot do any validations with ActiveStorage? That's a bummer. I am in the process of migrating from Shrine because I like that with AS I don't have to alter the structure of the models. I thought I could use this gem for validations but I am direct-uploading to DigitalOcean. With Shrine I had some validations but admittedly I wasn't doing direct uploads, so I don't know if there is the same issue with Shrine (@janko is there?)

Any workaround? I am only interested in validating that the file is an image between a min and a max file size.

igorkasyanchuk:
This gem doesn't do any validations using JS. So for direct uploads, it doesn't work

vitobotta:
Sorry for the confusion, I wasn't referring to js validations (I use Uppy for that) but server side validations. I got the file size validation working, it was a silly mistake from my part. Thanks for the gem!

igorkasyanchuk:
closing this since it's too old, and JS validations is out of scope of this gem

PR #168, opened 2022-10-09
Add processable image validator

kukicola:
Content type validation isn't enough to ensure that what users upload is an image and can be manipulated by MiniMagick or Vips. So my idea is to add additional validator which can check if a file can be processed by the library. 

For simplicity, I used `Metadata` class which has necessary methods. It probably should be extracted to separate class but I didn't want to make a huge refactor along the way

It should solve the following issues:
https://github.com/igorkasyanchuk/active_storage_validations/issues/91
https://github.com/igorkasyanchuk/active_storage_validations/issues/62
https://github.com/igorkasyanchuk/active_storage_validations/issues/130 (partially, since it's only for images, for all other files we can extend ContentTypeValidator to read a file and use `Marcel::MimeType.for`)

### DIFF
--- a/test/active_storage_validations_test.rb
+++ b/test/active_storage_validations_test.rb
@@ -271,6 +271,19 @@ class ActiveStorageValidations::Test < ActiveSupport::TestCase
     e.proc_image.attach(pdf_file)
     assert !e.valid?
     assert e.errors.full_messages.include?("Image has an invalid content type")
+
+    e = OnlyImage.new
+    e.image.attach(image_1920x1080_file)
+    e.proc_image.attach(image_1920x1080_file)
+    e.another_image.attach(tar_file_with_image_content_type)
+    assert !e.valid?
+    assert_equal e.errors.full_messages, ["Another image is not a valid image"]
+
+    e = OnlyImage.new
+    e.image.attach(image_1920x1080_file)
+    e.proc_image.attach(image_1920x1080_file)
+    e.any_image.attach(tar_file_with_image_content_type)
+    assert e.valid?
   rescue Exception => ex
     puts ex.message
     puts ex.backtrace.take(20).join("\n")
@@ -555,4 +568,8 @@ end
 
 def tar_file
   { io: File.open(Rails.root.join('public', '404.html.tar')), filename: '404.html.tar', content_type: 'application/x-tar' }
+end
+
+def tar_file_with_image_content_type
+  { io: File.open(Rails.root.join('public', '404.html.tar')), filename: '404.png', content_type: 'image/png' }
 end
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
 - master
-  ...
+  - add ProcessableImageValidator
 
 - 1.0.0
   - allow Procs as an input for validation options https://github.com/igorkasyanchuk/active_storage_validations/pull/135/files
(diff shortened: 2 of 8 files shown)


igorkasyanchuk:
@gr8bit could you please help me with a review? Difficult to focus on this right now because of current situation in Ukraine

gr8bit:
Hi @kukicola, I dealt with this exact subject (I think) a few days ago and I'd like your opinion on this issue: https://github.com/rails/marcel/issues/76
I overloaded the `Marcel::MimeType.for` method in an initializer for this purpose, changing the one line from the ticket above.

kukicola:
@gr8bit I'm not sure if we should call it a Marcel issue. Those fallbacks are probably there for some reason. If it was up to me I would suggest making `for_data` public (to keep current `for` behavior) and using it in ActiveStorage instead. 

Anyway, I would treat that as a separate thing. Minetype validation is still not enough here. Consider a file with magicbytes matching PNG file but the rest of the file is garbage (corrupted image, random bytes, malicious code, whatever). Validation should also fail in this case. That's why calling `image.avg` (vips) or image.valid? (minimagick) is important here - to check if file can be processed by library

gr8bit:
@kukicola Thank you for the feedback, I'll incorporate that in the issue.
As for this PR, what I understood: you want to make sure an attachment really is a "processable image" for Vips/ImageMagick, so for types they cannot handle (might be webp or jpeg xl depending on your processor) they should fail validation. Right?

kukicola:
@gr8bit Yes but it can be used also just to check if the image is safe to display somewhere (even without processing). I agree, that `real` is not a perfect name, I was working on it late in the evening and didn't have better ideas :). What about `valid_image` to avoid suggesting that user needs to process the image?

gr8bit:
> @gr8bit Yes but it can be used also just to check if the image is safe to display somewhere (even without processing). I agree, that `real` is not a perfect name, I was working on it late in the evening and didn't have better ideas :). What about `valid_image` to avoid suggesting that user needs to process the image?

I fully understand that. :D My point of view on it is: you cannot know if the user can actually display it - it might be something that ImageMagick can handle but the user's browser (or "application") cannot (some new formats or even aged ones like Ghostscript). So I was thinking: the only thing we can ensure using Vips/IM is: they can handle it, that's why think processable might be a good fit.
In terms of "valid": all images that we can process are definitely valid - but not all valid images we can definitely process. ;)

kukicola:
ok I agree, I'll update PR to "processable" later today

kukicola:
@gr8bit done, please take a look

test/active_storage_validations_test.rb
```diff
@@ -271,6 +271,13 @@ class ActiveStorageValidations::Test < ActiveSupport::TestCase
     e.proc_image.attach(pdf_file)
     assert !e.valid?
     assert e.errors.full_messages.include?("Image has an invalid content type")
+
+    e = OnlyImage.new
+    e.image.attach(image_1920x1080_file)
+    e.proc_image.attach(image_1920x1080_file)
+    e.another_image.attach(tar_file_with_image_content_type)
```
gr8bit, line 278: I think we should add a negative test here as well, ensuring that a `processable_image: false` will accept the .tar file as png as suggested by file extension and type.

gr8bit (changes requested):

gr8bit:
> @gr8bit done, please take a look

You're fast, I tried my best to keep up. ;)

kukicola:
I have a lot of time today :) fixes done

gr8bit:
@igorkasyanchuk everything looks fine in my opinion, this PR could be merged.

gr8bit (approved):

igorkasyanchuk:
very good, I have checked your discussion, and to me all looks good.

PR #169, opened 2022-10-10
Fix slowness

### LINKED ISSUE #127: Specs slowed down alot
Thanks for your gem, but you'd probably should know that after you added Marcel gem, it slowed specs a lot.
We use validations only in 2 places and every validation check took 5minutes locally and 12 minutes on GHActiions for each check.
I believe the reason for that is that Marcel (when you don't pass name) opens the file.
https://github.com/rails/marcel/blob/51a8f5535f927c1494bd6fc78abc3021aedf681b/lib/marcel/mime_type.rb#L64

version: 0.9.4

>     0.60988 seconds ./spec/models/model_spec.rb:15
>     0.54952 seconds ./spec/models/model_spec.rb:16
> 

version 0.9.5

>     314.05 seconds ./spec/models/model_spec.rb:15
>     311.79 seconds ./spec/models/model_spec.rb:16

sobrinho:
We don't need to specify either allowing and rejecting.

Fix #127

### DIFF
--- a/lib/active_storage_validations/matchers/content_type_validator_matcher.rb
+++ b/lib/active_storage_validations/matchers/content_type_validator_matcher.rb
@@ -92,16 +92,6 @@ module ActiveStorageValidations
         suffix = type.to_s.split('/').last
         { io: Tempfile.new('.'), filename: "test.#{suffix}", content_type: type }
       end
-
-      private
-
-      def content_type_keys
-        if Rails.gem_version < Gem::Version.new('6.1.0')
-          Mime::LOOKUP.keys
-        else
-          Marcel::TYPES.keys
-        end
-      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


gr8bit:
@sobrinho if I understand correctly, we still have a breaking behavior with this PR as @joshmcarthur pointed out in #127, right?

igorkasyanchuk:
@sobrinho this looks like a good solution. I think it worth to merge.

But can you please check if we need this method `content_type_keys`

image

sobrinho:
Yep, it is a breaking change in terms of not testing the other side of the validator.

gr8bit:
> Yep, it is a breaking change in terms of not testing the other side of the validator.

Ok, but I also think the advantages outweigh that. 👍

igorkasyanchuk:
@gr8bit from your side do you approve this PR?

gr8bit:
> @gr8bit from your side do you approve this PR?

Yes. 👍

mvz:
Why exactly is specifying both `allowing` and `rejecting` at the same time not allowed anymore? It looks like the rest of the code should handle that just fine.

mvz:
Thanks @sobrinho, I actually added a PR for that as well (#172) :slightly_smiling_face:

PR #172, opened 2022-10-20
Support using allowing and rejecting together on content type matcher

mvz:
The changes in version 1.0.1 blocked the option of specifying both `allowing` and `rejecting` on the content type matcher. It turns out this was not necessary for fixing the slowness, since that was caused by the default value of `rejecting` using the `Marcel` types.

This changes adds back the option of using `allowing` and `rejecting` together.

PR #173, opened 2022-10-20
Restore allowing and rejecting at the same time

sobrinho:


igorkasyanchuk:
@sobrinho I will close this PR, because we already have this https://github.com/igorkasyanchuk/active_storage_validations/pull/172/files

it has a bit more specs

But anyway thank you

igorkasyanchuk:
@sobrinho please check PR #172 (It looks exactly the same but just double check if I can merge)

sobrinho:
@igorkasyanchuk yep, #172 looks better 👍

PR #176, opened 2022-11-08
Support StringIO attachments

kukicola:
Hello!
I run into another issue. Take a look at the code copied from my app:

```
downloaded_image = ::OpenURI.open_uri(some_url)
uri = ::URI.parse(some_url)

user.avatar.attach(io: downloaded_image, filename: File.basename(uri.path))
```

If a fetched file is smaller than 10kb (or something like that, I don't remember) `downloaded_image` will be a StringIO, not a Tempfile. ActiveStorage supports it so `active_support_validations` should too but it doesn't. It's raising `no implicit conversion of StringIO into String` error.

So to fix that I copy StringIO to Tempfile before pushing it to MiniMagick of Vips. I tried to make it work without additional file but MiniMagick creates it underneath anyway and Vips didn't cooperate :)

### DIFF
--- a/lib/active_storage_validations/metadata.rb
+++ b/lib/active_storage_validations/metadata.rb
@@ -81,10 +81,11 @@ module ActiveStorageValidations
                   MiniMagick::Image.new(tempfile.path)
                 end
       else
-        image = if image_processor == :vips && defined?(Vips) && Vips::get_suffixes.include?(File.extname(read_file_path).downcase)
-                  Vips::Image.new_from_file(read_file_path)
+        file_path = read_file_path
+        image = if image_processor == :vips && defined?(Vips) && Vips::get_suffixes.include?(File.extname(file_path).downcase)
+                  Vips::Image.new_from_file(file_path)
                 elsif defined?(MiniMagick)
-                  MiniMagick::Image.new(read_file_path)
+                  MiniMagick::Image.new(file_path)
                 end
       end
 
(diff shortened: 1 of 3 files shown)


motymichaely:
Any plans to merge this?

gr8bit:
I'll review it shortly, promised :)

lib/active_storage_validations/metadata.rb
```diff
@@ -125,7 +126,18 @@ def read_file_path
       when ActionDispatch::Http::UploadedFile, Rack::Test::UploadedFile
         file.path
       when Hash
-        File.open(file.fetch(:io)).path
+        io = file.fetch(:io)
+        if io.is_a?(StringIO)
+          tempfile = Tempfile.new([File.basename(file[:filename], '.*'), File.extname(file[:filename])])
```
igorkasyanchuk, line 131: I know we have it even above in the code, but maybe we can also check if it is worth doing.

So the question is if we need to `close!` Tempfile.

kukicola: That would be nice but I don't see any nice way to do that without instance variable. Also, file created here: https://github.com/igorkasyanchuk/active_storage_validations/blob/f707239b4b486f0f1c35be49342883eb6dafb6a8/lib/active_storage_validations/metadata.rb#L68 isn't closed as well.

motymichaely:
Any update on this? :)

lib/active_storage_validations/metadata.rb
motymichaely: Any update on this?

motymichaely:
Ping :)

motymichaely:
@gr8bit - any plans to merge this?

lib/active_storage_validations/metadata.rb
gr8bit: According to the ruby Tempfile documentation:
> When a Tempfile object is garbage collected, or when the Ruby interpreter exits, its associated temporary file is automatically deleted. This means that’s it’s unnecessary to explicitly delete a Tempfile after use, though it’s good practice to do so
So I it should be okay to just have it garbage collected.

PR #178, opened 2022-12-17
add detailed error message

loqimean:
Screenshot 2022-12-17 at 19 46 22

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,7 +1,7 @@
 en:
   errors:
     messages:
-      content_type_invalid: "has an invalid content type"
+      content_type_invalid: "has not allowed to upload %{content_type} files, allowed types: %{authorized_types}"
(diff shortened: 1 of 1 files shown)


loqimean:
> Screenshot 2022-12-17 at 19 46 22

So, I can do it and update another files to improve the message errors, what do you think? Is it good idea?

gr8bit:
Hi @loqimean, the standard set of error messages should be suitable for all users, that's why they do not (by default) show as many details as you suggested.
As you know (read in README.rb), you can change (override) the translations in your own project to fit your needs.
We should keep the messages simple and adaptable for all users.

loqimean:
> Hi @loqimean, the standard set of error messages should be suitable for all users, that's why they do not (by default) show as many details as you suggested. As you know (read in README.rb), you can change (override) the translations in your own project to fit your needs. We should keep the messages simple and adaptable for all users.

So, okay, I get it

PR #179, opened 2023-01-04
[Validations] Validations fail when using dimension: { min: 1080..1080 } syntax

### LINKED ISSUE #163: [Validations] Validations fail when using dimension: { min: 1080..1080 } syntax
Hi,

In the readme, it is mentioned that we can use a short syntax for validating a min width and a min height at the same time, this looks like:  

```ruby
validates :feed_attachment, dimension: { min: 1080..1080 }
```

However, using this syntax makes the validations fail with RSpec, if I use:
```ruby
it do
  is_expected.to validate_dimensions_of(:feed_attachment).width_min(1080).height_min(1080)
end
```

It will throw me an error saying:

```ruby
     Failure/Error: is_expected.to validate_dimensions_of(:feed_attachment).width_min(1080).height_min(1080)
     
       is expected to validate dimensions of feed_attachment
         width between 1080 and 
         height between 1080 and 
```

Whereas if I use this syntax:

```ruby
validates :feed_attachment, dimension: { width: { min: 1080 } }
validates :feed_attachment, dimension: { height: { min: 1080 } }
```

The previous test passes perfectly.

I could not find in the gem code why it acts this way but this looks like a bug to me, I'll keep the 2 validations for now even if it is 'dirty', keep in mind that it will display 2 error messages when your user fails to respect both width & height.

The 2 way to write these validations works the same in the app.

Mth0158:
Closes #163 

This PR has 2 commits :
- [[Bugfix] Correct test fail using 'dimension: { min: 1080..1080 }'](https://github.com/igorkasyanchuk/active_storage_validations/commit/8da49577bd9bcf8d7d0b23699f9e38065a65f33c) => fixes #163 
- [[DimensionValidator] Make semantic changes for better understanding](https://github.com/igorkasyanchuk/active_storage_validations/commit/810f8eb8f0a4f405fef263b9200d394c41f7022f) => makes semantic changes for better understanding

The second commit is optional, though I think it's a good thing to change some methods & calls for making contributing slightly easier

Open to discussion, I would like to contribute further to this amazing project that you made @igorkasyanchuk ;)

### DIFF
--- a/test/matchers/dimension_validator_matcher_test.rb
+++ b/test/matchers/dimension_validator_matcher_test.rb
@@ -143,4 +143,62 @@ class ActiveStorageValidations::Matchers::DimensionValidatorMatcher::Test < Acti
     matcher.width_min 800
     refute matcher.matches?(Project.new)
   end
+
+  # width_min and height_min combined
+  test 'both width_min and height_min on higher combined are a positive match' do
+    matcher = ActiveStorageValidations::Matchers::DimensionValidatorMatcher.new(:dimension_min)
+    matcher.width_min 800
+    matcher.height_min 600
+    assert matcher.matches?(Project)
+  end
+
+  test 'width_min less than lower and height_min on higher combined are a negative match' do
+    matcher = ActiveStorageValidations::Matchers::DimensionValidatorMatcher.new(:dimension_min)
+    matcher.width_min 700
+    matcher.height_min 600
+    refute matcher.matches?(Project)
+  end
+
+  test 'width_min on higher and height_min less than lower combined are a negative match' do
+    matcher = ActiveStorageValidations::Matchers::DimensionValidatorMatcher.new(:dimension_min)
+    matcher.width_min 800
+    matcher.height_min 500
+    refute matcher.matches?(Project)
+  end
+
+  test 'both width_min and height_min less than lower combined are a negative match' do
+    matcher = ActiveStorageValidations::Matchers::DimensionValidatorMatcher.new(:dimension_min)
+    matcher.width_min 700
+    matcher.height_min 500
+    refute matcher.matches?(Project)
+  end
+
+  # width_max and height_max combined
+  test 'both width_max and height_max on higher combined are a positive match' do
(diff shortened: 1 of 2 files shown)


gr8bit:
@Mth0158 if you synced with the current master the minutest problems should be gone. :)

Mth0158:
@gr8bit all good, I have rebased onto `master` with your changes, all tests pass now.

I have also added the `dimension: { max: xxx ... yyy }` part since the problem would be the same as `min`. Let me know if the 2 commits are okay.

I can also help on other subjects if needed!

test/matchers/dimension_validator_matcher_test.rb
```diff
@@ -143,4 +143,62 @@ class ActiveStorageValidations::Matchers::DimensionValidatorMatcher::Test < Acti
...
+    matcher.width_min 800
+    matcher.height_min 600
+    assert matcher.matches?(Project)
+  end
+
+  test 'width_min less than lower and height_min on higher combined are a negative match' do
+    matcher = ActiveStorageValidations::Matchers::DimensionValidatorMatcher.new(:dimension_min)
+    matcher.width_min 700
+    matcher.height_min 600
+    refute matcher.matches?(Project)
```
gr8bit, line 159: Just because I saw it above - shouldn't this be `Project.new`?

Mth0158: Well, I have followed the same test pattern as the other tests. Here we are testing the test matcher, if we pass a class (`Project`) to the `matches?` method, it will convert it to `Project.new` automatically in the method.

test/matchers/dimension_validator_matcher_test.rb
```diff
@@ -143,4 +143,62 @@ class ActiveStorageValidations::Matchers::DimensionValidatorMatcher::Test < Acti
     matcher.width_min 800
     refute matcher.matches?(Project.new)
```
gr8bit, line 144: Regarding my comment below, I was referring to this line. But you're right and we should probably pass the class instead of an object everywhere.

gr8bit:
@Mth0158 thank you for your contribution! 👍

PR #180, opened 2023-01-04
fix minitest yaml syntax for ruby versions

gr8bit:
Use strings for ruby versions so yaml will never interpret them numerically.

Ruby 3.2 was released recently so an error testing "Ruby 3" occurred - it was supposed to test against "Ruby 3.0" but used Ruby 3.1, now Ruby 3.2. This fixes the yaml file by using strings for ruby versions anywhere.


gr8bit:
Tests on ruby 3.0 fixed.

PR #183, opened 2023-01-10
[Global] Better size validator messages + Refactor custom model error method

Mth0158:
Closes #82 <br>

This PR has 3 commits:

- [[SizeValidator] Better error messages depending on context, and make test coverage better](https://github.com/igorkasyanchuk/active_storage_validations/commit/142d3824d8c13b9bfd9650d188cebf9d5ac0f400) => fixes https://github.com/igorkasyanchuk/active_storage_validations/issues/82 + handles every test cases for size validator

I used Google translate for the translations, unsure it's the best translations though they seem pretty close to the previous one so if think it's okay?

- [[Validators] Make semantic changes for better understanding](https://github.com/igorkasyanchuk/active_storage_validations/commit/e9449b5880cb0a7af2adf29d6db53d4f22babf49)=> makes semantic changes for better understanding

- [[Global] Refactor custom model error method into a module thus making .with_message matcher available on all validators](https://github.com/igorkasyanchuk/active_storage_validations/commit/2490ce40e9280b49db737e744403485f2aa3f0ee) => shares the add_error method across validators, thus making the code easier to maintain + enable .with_message matcher on all validators

The second and third commit are optional, though I think it's a pretty nice addition that makes the code simple to understand and maintain + it makes the .with_message matcher available on all validators.

Open to discussion @gr8bit & @igorkasyanchuk

PR #184, opened 2023-02-09
Add swedish translation

Fonsan:


gr8bit:
Jättebra, tusen tack! :) We need to wait for #183 to be merged though because it will change/add some translations. I will mark you in this issue once that is reviewed & merged.

Issue #170, opened 2022-10-11
"foobar is not a valid image" for valid image (error only in production)


marckohlbrugge:
I have the following validation in my model:

```ruby
  validates :input_image,
    attached: true,
    content_type: ["image/png", "image/jpeg"],
    dimension: {
      width: { min: 10, max: 5000 },
      height: { min: 10, max: 5000 }
    },
    size: { less_than: 25.megabytes }
```

When trying to attach an image that fits these criteria, it works fine on localhost (disk storage). But when trying the same image in production (using S3), I get the error `Input message is not a valid image`. Unfortunately it does not specify what's not valid about it.

Any suggestions on how to debug this?

I'm using `rails (7.0.4)`, and `active_storage_validations (0.9.8)`.

gr8bit:
Hm, my semi-educated guess is: you're either
a) using an asynchronous job runner in production but not locally which delays image analyzation and therefore the adding of the width and height metadata the dimension validator needs or
b) missing a Vips/ImageMagick processor for the uploaded files in question.

I'd try to reproduce the error using an own "development" S3 bucket locally.

marckohlbrugge:
> a) using an asynchronous job runner in production but not locally which delays image analyzation and therefore the adding of the width and height metadata the dimension validator needs

That was it!

I have removed the dimensions validation (and kept the rest), and now it works as expected. Thank you!

I guess I can use something like [`fastimage`](https://github.com/sdsykes/fastimage) in a `before_validation` callback to set the height and width. Perhaps also worth adding as an option to this gem, as I believe analyzing an attachment asynchronously is the default in Rails?

gr8bit:
There's a mechanism for this already somewhere in ActiveStorage, I remember seeing some logic like "was the picture analyzed yet? if not, do it right away." in the AS code a few months ago. I will dive back into that.

marckohlbrugge:
> There's a mechanism for this already somewhere in ActiveStorage, I remember seeing some logic like "was the picture analyzed yet? if not, do it right away." in the AS code a few months ago. I will dive back into that.

I think this gem by already do this. The problem might be my own code disabling analyzers altogether.

(So please don't spend time looking into this just for me.)

marckohlbrugge:
I'm running into a similar issue in a different codebase.

The validations work fine in development. But on production I always get `"[…] is not a valid image"`

[...]

PR #185, opened 2023-02-15
Support libvips < 8.8

iainbeeston:
If you try to validate images while using libvips 8.7 every image fails validation with an "is not a valid image" error. This was helpfully investigated by @marckohlbrugge in #170 and seems to be caused because libvips can't report which file types it supports in versions below 8.8.

I've changed the code so it checks to see whether libvips can report the file types. I've assumed that if we have libvips installed and it can't report the supported file types then we should still use libvips. This might mean we raise errors if we give vips a file it does not support but I can't see another (easy) way to support older versions of libvips.

### DIFF
--- a/lib/active_storage_validations/metadata.rb
+++ b/lib/active_storage_validations/metadata.rb
@@ -75,18 +75,10 @@ module ActiveStorageValidations
         tempfile.flush
         tempfile.rewind
 
-        image = if image_processor == :vips && defined?(Vips) && Vips::get_suffixes.include?(File.extname(tempfile.path).downcase)
-                  Vips::Image.new_from_file(tempfile.path)
-                elsif defined?(MiniMagick)
-                  MiniMagick::Image.new(tempfile.path)
-                end
+        image = new_image_from_path(tempfile.path)
       else
         file_path = read_file_path
-        image = if image_processor == :vips && defined?(Vips) && Vips::get_suffixes.include?(File.extname(file_path).downcase)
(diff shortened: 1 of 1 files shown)


marckohlbrugge:
+1 on this. It resolved the issue described in #170

adenta-grandstand:
Any updates?

igorkasyanchuk:
Need help for this. If you or anyone can work on this it would be great

iainbeeston:
@igorkasyanchuk what help do you need? Are there any changes needed to the code in this PR before it can be merged?

igorkasyanchuk:
@iainbeeston yes, help to review PR and test if all is good, and it's ready to merge. Just need to 👀 on this

iainbeeston:
Maybe @marckohlbrugge @adenta-grandstand could confirm if they have tried this, and review the code to see if they agree with the changes?

marckohlbrugge:
This did solve the issue for me and the code change seems straight forward.

I think the only downside is that for older versions of vips, when using an incompatible file extension, the code will now fail differently.

Maybe someone can test this to see the difference in behavior?

Either way, this PR seems like an improvement over the existing code.

igorkasyanchuk:
looks like libvips 8.8 was released long time ago, so I think we are fine. People can just upgrade it. Thanks for help, will release gem soon

igorkasyanchuk:
oh, meanwhile if anyone has time to help to review other PRs please help

Issue #197, opened 2023-07-27
Gem depends on active_model, but doesn't require it before using it


technicalpickles:
I'm adding this gem to our project, and found problems when using sidekiq's sidekiqswam:

```
[swarm] Preloading Bundler groups ["default"]
/Users/josh.nichols/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/active_storage_validations-1.0.4/lib/active_storage_validations/attached_validator.rb:4:in `<module:ActiveStorageValidations>': uninitialized constant ActiveStorageValidations::ActiveModel (NameError)

  class AttachedValidator < ActiveModel::EachValidator # :nodoc:
                                       ^^^^^^^^^^^^^^^
Did you mean?  ActiveSupport
	from /Users/josh.nichols/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/active_storage_validations-1.0.4/lib/active_storage_validations/attached_validator.rb:3:in `<top (required)>'
	from <internal:/Users/josh.nichols/.rbenv/versions/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from <internal:/Users/josh.nichols/.rbenv/versions/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from /Users/josh.nichols/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/active_storage_validations-1.0.4/lib/active_storage_validations.rb:6:in `<top (required)>'
	from /Users/josh.nichols/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/bundler-2.4.17/lib/bundler/runtime.rb:60:in `require'
	from /Users/josh.nichols/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/bundler-2.4.17/lib/bundler/runtime.rb:60:in `block (2 levels) in require'
	from /Users/josh.nichols/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/bundler-2.4.17/lib/bundler/runtime.rb:55:in `each'
	from /Users/josh.nichols/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/bundler-2.4.17/lib/bundler/runtime.rb:55:in `block in require'
[...]

PR #198, opened 2023-07-27
Explicitly require active_model before using it

technicalpickles:
There are some cases where rails hasn't been fully initialized by the time active_storage_validations is required. In those cases, there are active_model classes in this library that are used before `active_model` has been required.

This fixes it by requiring it in the main file.

fixes https://github.com/igorkasyanchuk/active_storage_validations/issues/197


igorkasyanchuk:
thanks!

Issue #201, opened 2023-10-25
Existing specs started failing after upgrade to 1.1.0. Probably an issue with the matcher


tagliala:
Hi,

some existing specs started failing after upgrade to 1.1.0.

This is an issue on both Rails 7.0 and 7.1 with different Ruby versions

This has been bisected to 927ffdf (PR #183)

Reproducible test case:

```rb
# frozen_string_literal: true

class User < ApplicationRecord
  has_one_attached :avatar

  # Remove `size: { greater_than: 0 }` to prevent failure or downgrade to 1.0.4
  validates :avatar, content_type: %w[image/jpeg], size: { greater_than: 0 }
end
```

Spec:
```rb
RSpec.describe User do
  it { is_expected.to validate_content_type_of(:avatar).allowing('image/jpeg') }
end
```

#### 1.0.4
```
User
  is expected to validate the content types allowed on attachment avatar

Finished in 0.20146 seconds (files took 1.9 seconds to load)
1 example, 0 failures
```

#### 1.1.0
```
  1) User is expected to validate the content types allowed on attachment avatar
     Failure/Error: it { is_expected.to validate_content_type_of(:avatar).allowing('image/jpeg') }
     
       Expected avatar
       Accept content types: image/jpeg
        were rejected

Finished in 0.23732 seconds (files took 3.16 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/models/user_spec.rb:6 # User is expected to validate the content types allowed on attachment avatar
```

Repo: https://github.com/diowa/ruby3-rails7-bootstrap-heroku/tree/experimental/active-storage-validations

---

It appears to be a problem only in specs, the actual validation seems to work

igorkasyanchuk:
@Mth0158 can you check this please

Mth0158:
@igorkasyanchuk I'll look at it this afternoon and push a PR on this matter asap

tagliala:
Thanks for the fix and the prompt release

Alagaesia93:
Hi there!

I got into the same issue today and 1.1.1 does not fix it if you have a custom message. 

In your example
```ruby
  validates :avatar, content_type: %w[image/jpeg], size: { greater_than: 0, message: "whatever foobar" }
```

The spec fails

My example is with less_than, but I think it's the same

```ruby
class Shop < ApplicationRecord
  has_one_attached :logo
[...]

PR #206, opened 2023-10-27
203 validator custom message error causes test matcher error

### LINKED ISSUE #203: Validator custom error message causes test matcher error
From @Alagaesia93 in #201

---

Hi there!

I got into the same issue today and 1.1.1 does not fix it if you have a custom message. 

In your example
```ruby
  validates :avatar, content_type: %w[image/jpeg], size: { greater_than: 0, message: "whatever foobar" }
```

The spec fails

My example is with less_than, but I think it's the same

```ruby
class Shop < ApplicationRecord
  has_one_attached :logo
  validates :logo, size: { less_than: 10.megabytes }, content_type: ['image/png', 'image/jpg', 'image/jpeg'] # works
  validates :logo, size: { less_than: 10.megabytes, message: "must be less than 10 MB" }, content_type: ['image/png', 'image/jpg', 'image/jpeg'] # fails
end
```

```ruby
describe Shop do
  it { is_expected.to validate_size_of(:logo).less_than(10.megabytes) }
end
```

---

_Originally posted by @Alagaesia93 in https://github.com/igorkasyanchuk/active_storage_validations/issues/201#issuecomment-1780681426_

Mth0158:
@igorkasyanchuk here is a refacto proposition to make tests clearer.

In my opinion, switching to spec makes it really easier to read the tests, plus a big part of the community is using Rspec as their testing framework, so writing tests like specs makes even more sense.

I have just done the refacto for the size_validator_matcher as a POC, let me know if that's okay with you. Then, I was thinking of proceeding the following way:
- Refacto the whole gem tests => You validate
- Add new tests to mitigate regressions & fix #203 => You validate
Then we could merge it

Let me know how I can be helpful

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    active_storage_validations (1.1.0)
+    active_storage_validations (1.1.1)
       activejob (>= 5.2.0)
       activemodel (>= 5.2.0)
       activestorage (>= 5.2.0)
@@ -72,6 +72,8 @@ GEM
     mini_magick (4.11.0)
     mini_portile2 (2.5.1)
     minitest (5.14.3)
+    minitest-focus (1.4.0)
+      minitest (>= 4, < 6)
     nokogiri (1.11.3)
       mini_portile2 (~> 2.5.0)
       racc (~> 1.4)
@@ -138,6 +140,7 @@ DEPENDENCIES
   globalid
   marcel
   mini_magick (>= 4.9.5)
+  minitest-focus (~> 1.4)
   pry
   rubocop
   ruby-vips (>= 2.1.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,6 +46,7 @@ GEM
       zeitwerk (~> 2.3)
     ast (2.4.1)
     builder (3.2.4)
+    byebug (11.1.3)
     coderay (1.1.3)
     combustion (1.3.1)
       activesupport (>= 3.0.0)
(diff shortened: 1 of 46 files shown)


igorkasyanchuk:
@Mth0158 I'm good with this refactoring, please continue 🚀

Mth0158:
@igorkasyanchuk I think I'll:
- finish refactoring the 2 remaining matcher files (`attached` & `content_type`)
- add test for combined matchers (ie solving the #203 issue)

And maybe we can merge it if that's okay with you?
I will refactor the tests on the validators in another PR, since this PR is already adding a lot (as of now, it makes the test suite up to 360 from 232).

igorkasyanchuk:
@Mth0158 I'm fine with it. If you want to merge - do it.

Mth0158:
@igorkasyanchuk & @gr8bit I finished this PR.
I first added a lot of tests to make sure the big refacto in the last commit does not cause regressions.
The last commit is all about refactoring to simplify and DRY the code, it also fixes the issue #201 and add new inside features that can be reused somewhere else (the Validatable module, the Symobolizable module especially).
Let me know what you think of this, maybe it's better to only focus on the last commit since the other ones are all about cleaning and updating the test suite.

Mth0158:
I still think we could go further in refactoring the code, but I'll stop here for this PR, it's big enough (make too big actually) 😅

igorkasyanchuk:
@Mth0158 merged, thanks for the tremendous work. Do you think it worth to release a new version, or we can just wait.

Mth0158:
@igorkasyanchuk thanks for the merge!
Maybe we can release it as a minor version so it definitely fixes the bug mentioned in #203 & #201 that some users had. I do not know exactly how you plan releases so let me know what you think is best in this use case.
It's not a critical release anyway, more something that will be useful to work with going on

PR #209, opened 2023-10-31
Fix: Fix issue with incorrect error messages

jozzi05:
This PR fixes issues with randomly generated error codes when using extra options like `:if`, `:on` etc. Additionally, it adds a [new extension ](https://github.com/igorkasyanchuk/active_storage_validations/pull/209/files#diff-12b6445992a4e7a394a88f2216ca588eb9111f3af8f8fbd28f8b32bc15ebb3a5R52-R54)to the size matcher that enables users to define custom validation context.

#### Error messages: 
1. `on` example
* before:
`Size with context translation missing: en.activerecord.errors.models.size/portfolio.attributes.size_with_context.file_size_not_on`
* after:
`Size with context file size must be less than 2 KB (current size is 10 KB)`
2.  `if` example
* before:
`Size with if translation missing: en.activerecord.errors.models.size/portfolio.attributes.size_with_if.file_size_not_if`
* after:
`Size with if file size must be less than 2 KB (current size is 10 KB)`

PR #210, opened 2023-11-02
Fix: Fix size matcher for `has_many_attached` attachments

jozzi05:
This PR fixes the size validator matcher for `has_many_attached` attachments.

### Details:

Currently, the matcher checks all edge cases for size validation in chain:https://github.com/igorkasyanchuk/active_storage_validations/blob/dc765443e052a04660ab4cd17ac7f05ff9715cb8/lib/active_storage_validations/matchers/size_validator_matcher.rb#L54
this works for `has_one_attached` attachments since whenever `attach` method is called, previous attachment is being replaced with a new one. https://github.com/igorkasyanchuk/active_storage_validations/blob/dc765443e052a04660ab4cd17ac7f05ff9715cb8/lib/active_storage_validations/matchers/size_validator_matcher.rb#L92
Unfortunately, this is not the case for `has_many_attached` attachments since the `attach` method adds new attachments to the list, which leads to having error validations for attachments from previous checks. To solve this issue, I added cleanup of attachments after each step.
```ruby
.tap { @subject.attachment_changes["#{@attribute_name}"] = nil }
```

It works, but if there's a better way of fixing it, please let me know, and I will try to provide a better solution.

### DIFF
--- a/lib/active_storage_validations/matchers/size_validator_matcher.rb
+++ b/lib/active_storage_validations/matchers/size_validator_matcher.rb
@@ -92,9 +92,9 @@ module ActiveStorageValidations
           @subject.public_send(@attribute_name).attach(io: io, filename: 'test.png', content_type: 'image/pg')
           @subject.validate
           exclude_error_message = @custom_message || "file_size_not_"
-          @subject.errors.details[@attribute_name].none? do |error|
-            error[:error].to_s.include?(exclude_error_message)
-          end
+          @subject.errors.details[@attribute_name]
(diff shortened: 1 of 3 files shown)


jozzi05:
@igorkasyanchuk @Mth0158 I would love to hear your feedback also in regards to this fix 🙏

jozzi05:
Same as mentioned in other PR (https://github.com/igorkasyanchuk/active_storage_validations/pull/209#issuecomment-1805426141)

PR #214, opened 2023-11-12
213 incorrect error message when validator is used with if or on

Mth0158:
Thanks to the work done in #209 

Again, a "big" one, but it's mostly refacto the validator test suite. Only 2 prod files are impacted:
- `lib/active_storage_validations/limit_validator.rb` => We do not take into account the attached files without blob when counting for validating, @igorkasyanchuk I had to do this to pass the tests with context option, I think this was an edge case? Or is this something that I get wrong?
- `lib/active_storage_validations/size_validator.rb` => We make sure not to take Rails options as key interpolated for i18n


Mth0158:
It's in draft because it's missing the `if` option, I'll finish this part soon!

Mth0158:
@igorkasyanchuk I took the liberty to merge this PR since it only impacts lightly 2 prod files. This PR sets the base for the next test files that will greatly enhance our ability to mitigate regressions :)
Let me know if I overstepped, I'll let you do the release

Issue #216, opened 2023-11-15
1.1.2 release is breaking


aditya-cherukuri:
I believe the Validatable concern `active_storage_validations/matchers/concerns/validatable.rb` hasn't been namespaced correctly and hence is causing a errors when trying to run specs. Here is the stack trace,

```
/vendor/bundle/ruby/3.1.0/gems/active_storage_validations-1.1.2/lib/active_storage_validations/matchers/concerns/validatable.rb:2:in `<module:Validatable>': uninitialized constant Validatable::ActiveSupport (NameError)

  extend ActiveSupport::Concern
         ^^^^^^^^^^^^^
	from vendor/bundle/ruby/3.1.0/gems/active_storage_validations-1.1.2/lib/active_storage_validations/matchers/concerns/validatable.rb:1:in `<top (required)>'
	from vendor/bundle/ruby/3.1.0/gems/active_storage_validations-1.1.2/lib/active_storage_validations/matchers/attached_validator_matcher.rb:3:in `require_relative'
	from vendor/bundle/ruby/3.1.0/gems/active_storage_validations-1.1.2/lib/active_storage_validations/matchers/attached_validator_matcher.rb:3:in `<top (required)>'
	from vendor/bundle/ruby/3.1.0/gems/active_storage_validations-1.1.2/lib/active_storage_validations/matchers.rb:3:in `require'
	from vendor/bundle/ruby/3.1.0/gems/active_storage_validations-1.1.2/lib/active_storage_validations/matchers.rb:3:in `<top (required)>'
```

Mth0158:
Hi @aditya-cherukuri,
This bug has been fixed in your #217 PR and has been released today in https://github.com/igorkasyanchuk/active_storage_validations/commit/270212653bc775600caa79b908ffb6d83759ae40.
Thanks again for your work!

PR #217, opened 2023-11-16
[fix] validatable resoultion

aditya-cherukuri:
Fix for https://github.com/igorkasyanchuk/active_storage_validations/issues/216

### DIFF
--- a/lib/active_storage_validations/matchers/concerns/validatable.rb
+++ b/lib/active_storage_validations/matchers/concerns/validatable.rb
@@ -1,46 +1,53 @@
-module Validatable
-  extend ActiveSupport::Concern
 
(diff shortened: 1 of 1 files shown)


igorkasyanchuk:
@Mth0158 this look like a valid fix? Do you have any comments

Mth0158:
Hi
@igorkasyanchuk it looks okay to me, the only thing that I am troubled by is that we did not see it with the tests, how could we improve our test suite to catch this bug for the next releases?
@aditya-cherukuri thanks again for the quick PR and sorry for the short inconvenience

Mth0158:
@igorkasyanchuk can we merge and release this  PR?

PR #218, opened 2023-11-20
196 rails default validation options missing from errors

### LINKED ISSUE #196: Rails default validation options missing from errors
Hello,

I noticed that the [Rails' validation options](https://api.rubyonrails.org/classes/ActiveModel/Validations/ClassMethods.html#method-i-validates) (:if, :unless, :on, :allow_nil, :allow_blank, and :strict) are not added to the generated errors unlike Rails does with its own validators:

```ruby
class User < ApplicationRecord
  has_one_attached :avatar

  validates :avatar, attached: true, on: :foo
  validates :email, presence: true, on: :bar
end

user = User.new

user.valid?
=> true

user.valid?(:bar)
=> false

user.errors
=> #<ActiveModel::Errors [#<ActiveModel::Error attribute=email, type=blank, options={:on=>:bar}>]> # 'on' option is present

user.valid?(:foo)
=> false

user.errors
=> #<ActiveModel::Errors [#<ActiveModel::Error attribute=avatar, type=blank, options={}>]> # 'on'  option is missing !
```

Is there any reason why these options are not added to the errors ? If not, it would be awesome to have them.

Thanks for the gem BTW 💎

Mth0158:
Closes #196

This PR includes 2 commits:
- ab567ad325ac82b14bef2535d9edfc7a24cab3a9: This commit have a lot a files changed because I had to create a lot of boilerplate to test it properly, it only impacts a few production features =>
  - The `error_handler.rb` file becomes the Errorable concern and gets a few upgrade to adress #196
  - The `attached` validator now does not accept the `:allow_nil` and `:allow_blank` options since it does not make sense to have these options with this validator
- 5a85ba7aff9246b3f61363b958cb45ee9b35c2e2: A short commit that defines our validator concerns at the gem level instead that the global level


Mth0158:
@igorkasyanchuk I would like to merge the #217 PR before this one though :)

PR #221, opened 2023-12-01
212 add filename to error message

Mth0158:


Mth0158:
@igorkasyanchuk FYI, we can proceed to the merge

Issue #234, opened 2023-12-13
Problem with conditional validations


sergiopatricio:
I have the following model:
```ruby
class Document < ApplicationRecord
  has_one_attached :asset

  validates :asset, attached: true

  validates :asset, size: { less_than: 20.megabytes, message: 'must be smaller than 20MB' }, if: :condition1?
  validates :asset, size: { less_than: 10.megabytes, message: 'must be smaller than 10MB' }, if: :condition2?
end
```

When I run this test:
```ruby
context 'when condition2' do
  subject(:document) { Document.new(...set condition2...) }

  it { is_expected.to validate_size_of(:asset).less_than(10.megabytes) }
end
```
the test fails.

I noticed that if I remove `message` from the validator options the test passes.

This problem started happening when updating to `>= 1.1.0`.

---
I did some debugging and the problem seems to be in https://github.com/igorkasyanchuk/active_storage_validations/blob/master/lib/active_storage_validations/matchers/concerns/validatable.rb, with this code:
```ruby
@subject.class.validators_on(@attribute_name).find do |validator|
  validator.class == validator_class
end
```
This will use the first `SizeValidator` that it finds, but it may not be the correct one.

```ruby
 @subject.class.validators_on(@attribute_name)
=> [#<ActiveStorageValidations::AttachedValidator:0x000000010efb5388 @attributes=[:asset], @options={}>,
 #<ActiveStorageValidations::SizeValidator:0x000000010ef7f4b8 @attributes=[:asset], @options={:if=>:condition1?, :less_than=>20971520, :message=>"must be smaller than 20MB"}>,
 #<ActiveStorageValidations::SizeValidator:0x000000010ef7eba8 @attributes=[:asset], @options={:if=>:condition2?, :less_than=>10485760, :message=>"must be smaller than 10MB"}>]
```
It finds the `SizeValidator` for `20MB`, but I'm testing the validator for "10MB".

Mth0158:
Hi @sergiopatricio,
Thanks for raising the issue and the debugging, we are apparently missing an extra step to identify the attribute that raises the error.
A good fix would be to add another condition in the `Validatable` method that you mentioned, would you mind creating a PR with the fix + its associated tests? I would encourage the use of a shared_example for the `Validatable` concern that we could use on all validators during the test phase.

sergiopatricio:
Thanks @Mth0158, I can try although I need to investigate a bit more how I can create a test for this.

[...]

PR #236, opened 2024-01-02
[Matcher] Bugfix when several validator are used on the same attribute (#234)

Mth0158:


Mth0158:
It needs to handle `contextable` use case to be merged, it should follow soon

PR #241, opened 2024-02-08
Add install instructions to Readme

stevehanson:
I didn't see the gem name or install instructions listed in the Readme and had to open the gemspec to find the gem name. This adds install instructions at the top.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,18 @@ Add to your Gemfile:
 gem "active_storage_validations"
 ```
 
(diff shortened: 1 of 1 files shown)


Mth0158:
Hi @stevehanson 
Thks for your PR, but there is an instruction for installing the gem in your Rails app here: https://github.com/igorkasyanchuk/active_storage_validations?tab=readme-ov-file#installation#241

PR #244, opened 2024-03-04
Use 0 as byte_size if blob is nil

Sivard:


Mth0158:
Hi @Sivard 
Could you give us a bit of context for this change?

PR #247, opened 2024-06-03
Remove outdated version comment

wataori:
Hello. 
Thank you for your maintaining this repo. We've just started to use this nice validator and I noticed README was outdated. Rails 5.2 support seems ended at 1.0.0 and 6.0 was also ended at 1.1.3. I considered updating comment's version with latest 6.1 and 7 but I didn't because It's okay without it.


Mth0158:
Thanks @wataori!

PR #249, opened 2024-06-28
Accept content types like "image/*"

manuelmeurer:
I would like to be able to use content type validations like "image/*".

Currently, this is only supported with a regex (/\Aimage\/*\z/) which is kind of ugly and unnecessarily complicated IMO, but also has the downside that the value cannot directly be used as [the "accept" attribute on a HTML input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept), which is what I'm doing in my app.

It turns out that accepting strings like "image/*" was a relatively small change.

If this is an acceptable change to be merged, I am happy to add some tests for it as well! 😸

### DIFF
--- a/lib/active_storage_validations/content_type_validator.rb
+++ b/lib/active_storage_validations/content_type_validator.rb
@@ -60,7 +60,9 @@ module ActiveStorageValidations
     def is_valid?(file, types)
       file_type = content_type(file)
       types.any? do |type|
-        type == file_type || (type.is_a?(Regexp) && type.match?(file_type.to_s))
+        type == file_type ||
+          (type.is_a?(Regexp) && type.match?(file_type.to_s)) ||
+          (type.is_a?(String) && type.end_with?("/*") && file_type.start_with?(type[0..-2]))
       end
     end
 


Mth0158:
Hi @manuelmeurer 
Thanks for the PR. I am reviewing it, and I am unsure about adding multiple ways of doing the same thing with the gem. I prefer keeping things simple.
Couldn't you just do this change in your app? Converting `"image/*"` into the appropriate regex and vice-versa is easy.
Let's see what @igorkasyanchuk think of it :)

manuelmeurer:
Thanks for your feedback @Mth0158!

Matching the regex to "image/*" would be kinda ugly and error prone IMO...
Is "image/*" the same as /\Aimage\/*\z/? What about /\Aimage\/[a-z]+\z/ or /\mage\/*/?

To be honest, I don't think a regex is appropriate for this content type validation. Its flexibility is not really needed, the only use case is /\Aimage\/*\z/ (correct me if I'm wrong) and that can be expressed more succinctly with "image/*" (AND it would have the added benefit of being usable as the "accept" attribute).

gr8bit:
Just to add to the use cases of regexes, I found these examples in one of our apps, narrowing the allowed content types of specific uploads with the active storage validations:
```ruby
  %r{\Aimage/(jpe?g|png|hei[fc])\z} # specific set of image types
  %r{\Aapplication/vnd\.openxmlformats-officedocument\.} # all office documents
```

Regarding the conversion of the **accept**-attribute value `image/*` to a regex, it would be `/\Aimage//` ("start with `image/`").

My 2 cents on the "dual use" of the **accept**-attribute and the validation:
As the accept-attribute also allows to specify filename extensions like `.docx` and generally multiple, comma-separated types (which you would miss out on using if you mapped the content type directly), I'd suggest a simple mapping of "an image" to a) the value for the accept-attribute and b) the according allowed content types for validation.
Of course I don't know your use case, but does your application handle for example SVG images (mime type `image/svg+xml`), MS icons (`image/vnd.microsoft.icon`) or animated PNGs (`image/apng`) just like it handles jpeg? For me, they always were some of the stranger fellows lurking around in the `image/`-namespace. ;)

manuelmeurer:
> Of course I don't know your use case, but does your application handle for example SVG images (mime type image/svg+xml), MS icons (image/vnd.microsoft.icon) or animated PNGs (image/apng) just like it handles jpeg? For me, they always were some of the stranger fellows lurking around in the image/-namespace. ;)

That's a fair point, maybe it is better to be more restrictive... but then I have to keep an eye on "newer" image formats like WEBP, JPEG XL, JPEG 2000 and what have you, and add them to the list if they get popular.

> %r{\Aapplication/vnd\.openxmlformats-officedocument\.}

This might also be too permissive, does your app handle all possible office document formats? :)

> As the accept-attribute also allows to specify filename extensions like .docx and generally multiple, comma-separated types (which you would miss out on using if you mapped the content type directly),...

I handle multiple types like this:

```ruby
validator
  .options
  .then { _1[:in] || _1.fetch(:with) }
  .then { Array(_1).join(", ") }
```

> I'd suggest a simple mapping of "an image" to a) the value for the accept-attribute and b) the according allowed content types for validation.

Ok, maybe this is the cleanest approach, defining the concept of "this attachment accepts a common image format" and generating the "accept" attribute and content type validation from there.

manuelmeurer:
>  %r{\Aimage/(jpe?g|png|hei[fc])\z}

AFAIC there is no content type `image/jpg`, it's always `image/jpeg`, no matter whether the file extension is .jpg or .jpeg.
So you could remove the `?` from the regex! 😸

PR #250, opened 2024-07-03
has_many_attached validate_size_of fix

yenshirak:
`ActiveStorageValidations::Matchers::SizeValidatorMatcher#attach_file` is currently using `attach`, which adds a new attachment instead of replacing the existing one when the relation type is `has_many_attached`. This causes a false negative because the validation in `higher_than_min?` fails due to the invalid attachment that was previously created by `not_lower_than_min?`.

The proposed solution is to use `"#{@attribute_name}="`, which will clear all previously attached files.

### DIFF
--- a/lib/active_storage_validations/matchers/size_validator_matcher.rb
+++ b/lib/active_storage_validations/matchers/size_validator_matcher.rb
@@ -139,7 +139,11 @@ module ActiveStorageValidations
       end
 
       def attach_file
-        @subject.public_send(@attribute_name).attach(dummy_file)
+        @subject.public_send("#{@attribute_name}=", attachable)
+      end
+
+      def attachable
+        @subject.public_send(@attribute_name).class == ActiveStorage::Attached::Many ? [dummy_file] : dummy_file
       end
 
       def dummy_file
--- a/test/dummy/app/models/size/matcher.rb
+++ b/test/dummy/app/models/size/matcher.rb
@@ -66,7 +66,7 @@ class Size::Matcher < ApplicationRecord
   has_one_attached :less_than_or_equal_to_with_message
   has_one_attached :greater_than_with_message
   has_one_attached :greater_than_or_equal_to_with_message
-  has_one_attached :between_with_message
+  has_many_attached :between_with_message
   validates :less_than_with_message, size: { less_than: 2.kilobytes, message: 'File is too big.' }
   validates :less_than_or_equal_to_with_message, size: { less_than_or_equal_to: 2.kilobytes, message: 'File is too big.' }
   validates :greater_than_with_message, size: { greater_than: 7.kilobytes, message: 'File is too small.' }


lib/active_storage_validations/matchers/size_validator_matcher.rb
```diff
@@ -111,8 +111,9 @@ def not_higher_than_max?
 
       def passes_validation_with_size(size)
         mock_size_for(io, size) do
-          attach_file
-          validate
+          attach_file do
+            validate
+          end
```
Mth0158, line 116: Could you do 
```ruby
attach_file
validate
remove_file
```
instead?

This fix logic can be replicated to other matchers in `has_many` case. Some can't use the `yield` logic like `AspectRatioValidatorMatcher`

test/matchers/size_validator_matcher_test.rb
```diff
@@ -88,10 +88,19 @@ module OnlyMatchWhenExactValue
...
+      let(:model_attribute) { matcher_method }
+
+      include SizeValidatorMatcherTest::OnlyMatchWhenExactValue
+    end
+
+    describe 'with has_many_attached' do
+      let(:model_attribute) { :many_greater_than }
+
+      include SizeValidatorMatcherTest::OnlyMatchWhenExactValue
+    end
```
Mth0158, line 103: Better do something like:
```ruby
    let(:model_attribute) { matcher_method }
    let(:validator_value) { 7.kilobytes }

    describe 'with has_one_attached' do
      let(:matcher_method) { :greater_than }

      include SizeValidatorMatcherTest::OnlyMatchWhenExactValue
    end

    describe 'with has_many_attached' do
      let(:matcher_method) { :many_greater_than }

      include SizeValidatorMatcherTest::OnlyMatchWhenExactValue
    end
```

Mth0158 (commented):
Hi @yenshirak 
Thanks for the PR, you are right, let fix this bug.
Plus, the logic can be replicated to other matchers, if you want to replicate it go for it, or I can do it :)
Whatever you prefer! Thanks again!

lib/active_storage_validations/matchers/size_validator_matcher.rb
yenshirak: done

test/matchers/size_validator_matcher_test.rb
yenshirak: I don't understand what you mean. Do you want me to implement a separate matcher method for `many_greater_than` in `SizeValidatorMatcher` which would be the same as `greater_than`?

yenshirak: Done. I had to create an alias for `SizeValidatorMatcher#many_greater_than`.

Mth0158: @yenshirak I think I made a mistake, my bad,your original code was the right one. We shouldntt have to add an alias in the `size_validator_matcher.rb` file.
Sorry for the inconvenience, could you revert this change?

lib/active_storage_validations/matchers/size_validator_matcher.rb
Mth0158: Thanks a lot for the concern, it's definitely the way to go 👍

test/matchers/size_validator_matcher_test.rb
yenshirak: done

Mth0158: Thanks a lot @yenshirak!

Issue #113, opened 2021-04-16
Validate total size of attachments in case of has_many relationship


a-moreira:
This is a great gem, thanks a lot.
As far as I understand, it's not possible to validate the total file size (summation of each file size) in case we have a `has_many` relationship. For example, I have the following contraint:

- upload some documents summing up to 10MB in size, each document with max 1MB each.

Currently I'm validating manually in the Model, but it'd be useful to have this in the gem as well.

I can try to implement it with some guidance.

Mth0158:
Hi @a-moreira,
Thanks for raising this issue.
It seems like a nice addition to the gem. I can guide you on how to implement it, or I can do it myself if you are too busy right now.
Let me know!

a-moreira:
hey @Mth0158, thanks for the reply. I'm busy right now so feel free to implement it :-)

mattiasgullstrand:
Hi! Any updates on this feature? @Mth0158

Mth0158:
Hi, @mattiasgullstrand I am busy these days, maybe I will find some time to implement it, but I can't say when...
It would be a nice addition to the gem though, feel free to assign it to yourself if you want to work on it :)

mattiasgullstrand:
Thanks @Mth0158. I see, and I agree, would be a nice addition! But not sure that I will have time anytime soon either

PR #252, opened 2024-07-04
[Validator/Matcher] Add new total_size validator and associated matcher (#113)

Mth0158:
- [x] Implement base validator
- [x] Add validator matcher
- [x] Implementing corresponding tests
- [x] Add translations
- [x] Update readme


Mth0158:
@igorkasyanchuk & @gr8bit what do you think of this PR?
As mentioned in #113, I have added a new validator to the gem: the `total_size`validator, which goal is to validate the total file size of a has_many_attached relation.
I have undergone a refacto of the `SizeValidator` / `SizeValidatorMatcher` classes in order to stay as DRY as possible, it feels ok to me. Hopefully we do not have regressions thanks to the test suite that is pretty dense now! :)

igorkasyanchuk (approved):

Issue #195, opened 2023-06-27
New validation option to check that file extension matches mime type


tagliala:
Hi,

I would like to propose a new validation that checks if the extension matches the actual content type of the document. This should help to prevent upload of a `.pdf` which is actually a `.docx`

I don't know a good name for this validation, or if it should be an option of `content_type`, like `match_extension: true`

I can work on a PR if you are interested

igorkasyanchuk:
Let's wait a few days for comment's. Could be useful.

valscion:
This sounds like content type spoofing validation. Similar to what `file_validators` has: require "rails"

%w(
  active_record/railtie
  active_storage/engine
  action_controller/railtie
  action_view/railtie
  action_mailer/railtie
  active_job/railtie
  action_cable/engine
  action_mailbox/engine
  action_text/engine
  rails/test_unit/railtie
  sprockets/railtie
).each do |railtie|
  begin
    require railtie
  rescue LoadError
  end
end

lavaturtle:
This sounds like a good idea to me too.

tagliala:
> NOTE: Also starting at version 4.0.0, Paperclip has another validation that cannot be turned off. This validation will prevent content type spoofing. That is, uploading a PHP document (for example) as part of the EXIF tags of a well-formed JPEG. This check is limited to the media type (the first part of the MIME type, so, 'text' in text/plain). This will prevent HTML documents from being uploaded as JPEGs, but will not prevent GIFs from being uploaded with a .jpg extension. This validation will only add validation errors to the form. It will not cause errors to be raised.

this is also interesting

Mth0158:
Yes that looks like a really good addition to the gem, I'll work on it in the coming days, let me know if you want to help :)

Mth0158:
Hi everyone! @tagliala, @valscion, @lavaturtle

I finally finished the PR about this feature, it was a not so easy task. I had to try several approaches to find the most suitable.
If anyone of you is interested in having a quick look at the PR before we merge it it would be appreciated :)

[https://github.com/igorkasyanchuk/active_storage_validations/pull/256](https://github.com/igorkasyanchuk/active_storage_validations/pull/256)

tagliala:
Hi, thanks for your hard work on the PR implementing this

There are a lot of changes, I would start backporting as many non-breaking changes as possible to keep the PR as small as possible

Also the change replacing OpenStruct with Struct is interesting (if it is only used there)

[...]

PR #256, opened 2024-07-29
[Validator] Add mime type spoofing detector (#195)

Mth0158:
This PR will require you to update to Rails 6.1.4 (was 6.1.3 before). This due to the fact that Marcel 0.3 is a dependy of Rails 6.1.3 and Marcel 1.0 is a dependency of Rails 6.1.4.

For this PR to work properly (ie accept closely related content type as a match), we need Marcel 1.0.


Mth0158:
@igorkasyanchuk & @gr8bit what do you think of this PR? :)

valscion:
Really nice work! I do agree that having this as a default is a sensible decision.

However, I'm not sure how many projects are currently relying on this gem already. This change in default functionality is quite big so I'd suggest shipping this change in two parts:

1. First in a 1.x release with the spoofing detection turned off
2. Then a new 2.x major release with the spoofing detection turned on

If you'd like to, you could also ignore doing a 1.x release and ship this directly to a 2.x release. I'd still encourage you to release this as a major version change as this could have subtle and surprising errors pop up for people's projects.

Mth0158:
@valscion thanks for taking time to answer!

> First in a 1.x release with the spoofing detection turned off
> Then a new 2.x major release with the spoofing detection turned on

I do agree with you, I will update the PR to reflect these changes. We will release a v2 with the spoofing protection as default.

igorkasyanchuk (approved):

MaksJS:
I've tried the new spoofing detection feature. Very nice work.

Unfortunately, I get an exception:

```
      NoMethodError:
        undefined method `attachable' for nil
```

when trying to update a record after creation. It's due to this line:

`@record.attachment_changes[@attribute.to_s].attachable`

After the first record save, when calling `update!` on the same record for example, `@record.attachment_changes` is empty (`{}`).

Issue #258, opened 2024-08-09
Add Danish translations


Mth0158:
The gem could use new translations. Danish could be a good one :)

kronn:
This issue has been fixed by #259 and could be closed, right?

Mth0158:
Yes! I am closing it now :)

PR #259, opened 2024-08-09
add danish translation

Scorpahr:
Add file dn.yml to have danish translation for errors validations

### DIFF
--- a/config/locales/dn.yml
+++ b/config/locales/dn.yml
@@ -1,0 +1,28 @@
+# Danish
+dn:
+  errors:
+    messages:
+      content_type_invalid: "har en ugyldig indholdstype"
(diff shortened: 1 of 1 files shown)


config/locales/dn.yml
```diff
@@ -0,0 +1,28 @@
+# Danish
+dn:
```
Mth0158, line 2: For Danish translations, it should use the Danish language code which is `da` (not `dn`).
Could you change this and the filename accordingly?

Mth0158 (changes requested):

Mth0158:
Closes #258

config/locales/dn.yml
Scorpahr: I change the name of the file and the accronyme in the file

Mth0158 (approved):

Mth0158:
@igorkasyanchuk I am merging this one :)

PR #267, opened 2024-10-11
Remove dead code for Rails 5.x, 6.x, and < 6.1.4

tagliala:
PR igorkasyanchuk#256 updated the minimum required version of Rails subcomponents to
6.1.4.

However, it also included conditionals specific to version 6.1.4.

This code is unnecessary as the library does not support versions of
Rails below 6.1.4.

---

Please review with "hide whitespace" option enabled

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -455,18 +455,6 @@ Tips:
 - To focus a specific file, use the TEST option provided by minitest, e.g. to only run size_validator_test.rb file you will execute the following command: `bundle exec rake test TEST=test/validators/size_validator_test.rb`
 
 
-## Known issues
-
-- There is an issue in Rails which it possible to get if you have added a validation and generating for example an image preview of attachments. It can be fixed with this:
-
-```erb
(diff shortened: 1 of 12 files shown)


Mth0158:
Hi @tagliala (again!)
I am totally on for these changes, I wanted to do this in the coming days/weeks, @igorkasyanchuk are you ok to deprecate support for Rails 5.x, and Rails 6 < 6.1.4?
I think it could be in the same release as the https://github.com/igorkasyanchuk/active_storage_validations/pull/256 release. Unsure if it's better when we drop support for a major of Rails to make a major or minor release?

tagliala:
Hi,

> are you ok to deprecate support for Rails 5.x, and Rails 6 < 6.1.4?

It has been already removed in current main:

https://github.com/igorkasyanchuk/active_storage_validations/blob/bd696b20d6ac9566e38225e2b34b8a38aec03739/active_storage_validations.gemspec#L25

PR #289, opened 2024-11-12
Support file extensions in uppercase

matthieuprat:
Since version 1.3.1 (fa64570), the extension of an attachment is validated against its content type. However, when the extension is not in lowercase, validation fails.

This commit fixes the issue by downcasing the attachment's extension before checking it against possible extensions for the content type.

### DIFF
--- a/test/dummy/app/models/content_type/validator/check.rb
+++ b/test/dummy/app/models/content_type/validator/check.rb
@@ -27,6 +27,8 @@ class ContentType::Validator::Check < ApplicationRecord
   validates :extension_two_extensions_docx, content_type: 'application/vnd.openxmlformats-officedocument.wordprocessingml.document'
   has_one_attached :extension_two_extensions_pdf
   validates :extension_two_extensions_pdf, content_type: 'application/pdf'
+  has_one_attached :pdf_file
+  validates :pdf_file, content_type: 'application/pdf'
 
   %w(symbol string regex).each do |type|
     has_one_attached :"with_#{type}"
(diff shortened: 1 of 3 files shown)


test/dummy/app/models/content_type/validator/check.rb
```diff
@@ -27,6 +27,8 @@ def self.example_for(type, several: false)
   validates :extension_two_extensions_docx, content_type: 'application/vnd.openxmlformats-officedocument.wordprocessingml.document'
   has_one_attached :extension_two_extensions_pdf
   validates :extension_two_extensions_pdf, content_type: 'application/pdf'
+  has_one_attached :pdf_file
+  validates :pdf_file, content_type: 'application/pdf'
```
Mth0158, line 31: To be consistent with the attribute naming, could you use `extension_upcase_extension` and `extension_missing_extension` as attribute names?
The change should also be reflected in `test/validators/content_type_validator_test.rb`

Mth0158 (changes requested):
Thanks a lot for the PR!
That's a case we didn't handle, thanks for adding the tests also :)

test/dummy/app/models/content_type/validator/check.rb
Mth0158: @matthieuprat

matthieuprat: Done!

Mth0158: Thanks a lot, all good to me!

PR #294, opened 2024-11-13
Allow multiple aspect ratios

### LINKED ISSUE #230: Add support for multiple aspect ratios
Like content_type, we should be able to support multiple aspect ratios.

Possible to get inspiration from this fork: https://github.com/igorkasyanchuk/active_storage_validations/compare/master...nejdetkadir:active_storage_validations:master. But not too much because they modified the :with option, we would rather like something with :in like content_type validator does.

Do not forget to add tests to your validator update + update the associated matcher

Please notify the fork owner so he/she can update its gemfile

mohammednasser-32:
Fixes #230 

Add support for validating multiple aspect ratios, as done for content types, like:
```ruby
validates :avatar, aspect_ratio: [:square, :portrait, :is_16_9]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ en:
       aspect_ratio_not_landscape: "must be a landscape image"
       aspect_ratio_is_not: "must have an aspect ratio of %{aspect_ratio}"
       image_not_processable: "is not a valid image"
+      aspect_ratio_invalid: "has invalid aspect ratio"
 ```
 
 In several cases, Active Storage Validations provides variables to help you customize messages:
--- a/config/locales/da.yml
+++ b/config/locales/da.yml
@@ -30,3 +30,4 @@ da:
       aspect_ratio_not_landscape: "skal være et landskabsbillede"
       aspect_ratio_is_not: "skal have et størrelsesforhold på %{aspect_ratio}"
       image_not_processable: "er ikke et gyldigt billede"
+      aspect_ratio_invalid: "har et ugyldigt billedformat"
--- a/lib/active_storage_validations/aspect_ratio_validator.rb
+++ b/lib/active_storage_validations/aspect_ratio_validator.rb
@@ -37,24 +37,18 @@ module ActiveStorageValidations
     def validate_each(record, attribute, _value)
       return if no_attachments?(record, attribute)
 
+      flat_options = set_flat_options(record)
+      @authorized_aspect_ratios = authorized_aspect_ratios_from_options(flat_options).compact
+      return if @authorized_aspect_ratios.empty?
+
       validate_changed_files_from_metadata(record, attribute)
     end
 
     private
 
     def is_valid?(record, attribute, attachable, metadata)
-      flat_options = set_flat_options(record)
-
-      return if image_metadata_missing?(record, attribute, attachable, flat_options, metadata)
-
-      @authorized_aspect_ratios = authorized_aspect_ratios(flat_options).compact
-      
-      return true if authorized_aspect_ratio?(record, attribute, attachable, metadata)
-
-      errors_options = initialize_error_options(options, attachable)
-      errors_options[:aspect_ratio] = string_aspect_ratios(flat_options)
-      add_error(record, attribute, aspect_ratio_error_mapping, **errors_options)
-      false
+      !image_metadata_missing?(record, attribute, attachable, metadata) &&
+        authorized_aspect_ratio?(record, attribute, attachable, metadata)
     end
 
     def authorized_aspect_ratio?(record, attribute, attachable, metadata)
@@ -66,20 +60,27 @@ module ActiveStorageValidations
         when ASPECT_RATIO_REGEX then valid_regex_aspect_ratio?(authorized_aspect_ratio, metadata)
         end
       end
+
+      return true if attachable_aspect_ratio_is_authorized
+
+      errors_options = initialize_error_options(options, attachable)
+      errors_options[:aspect_ratio] = string_aspect_ratios
+      add_error(record, attribute, aspect_ratio_error_mapping, **errors_options)
+      false
     end
 
     def aspect_ratio_error_mapping
-      return :aspect_ratio_invalid unless @authorized_aspect_ratios.length == 1
+      return :aspect_ratio_invalid unless @authorized_aspect_ratios.one?
 
       aspect_ratio = @authorized_aspect_ratios.first
       NAMED_ASPECT_RATIOS.include?(aspect_ratio) ? :"aspect_ratio_not_#{aspect_ratio}" : :aspect_ratio_is_not
     end
 
-    def image_metadata_missing?(record, attribute, attachable, flat_options, metadata)
+    def image_metadata_missing?(record, attribute, attachable, metadata)
       return false if metadata[:width].to_i > 0 && metadata[:height].to_i > 0
 
       errors_options = initialize_error_options(options, attachable)
-      errors_options[:aspect_ratio] = string_aspect_ratios(flat_options)
+      errors_options[:aspect_ratio] = string_aspect_ratios
       add_error(record, attribute, :image_metadata_missing, **errors_options)
       true
     end
@@ -113,7 +114,7 @@ module ActiveStorageValidations
     def ensure_aspect_ratio_validity
       return true if options[:with]&.is_a?(Proc) || options[:in]&.is_a?(Proc)
 
-      authorized_aspect_ratios(options).each do |aspect_ratio|
+      authorized_aspect_ratios_from_options(options).each do |aspect_ratio|
         unless NAMED_ASPECT_RATIOS.include?(aspect_ratio) || aspect_ratio =~ ASPECT_RATIO_REGEX
           raise ArgumentError, invalid_aspect_ratio_message
         end
@@ -128,12 +129,12 @@ module ActiveStorageValidations
       ERROR_MESSAGE
     end
 
-    def authorized_aspect_ratios(flat_options)
+    def authorized_aspect_ratios_from_options(flat_options)
       (Array.wrap(flat_options[:with]) + Array.wrap(flat_options[:in]))
     end
 
-    def string_aspect_ratios(flat_options)
-      authorized_aspect_ratios(flat_options).map do |aspect_ratio|
+    def string_aspect_ratios
+      @authorized_aspect_ratios.map do |aspect_ratio|
         if NAMED_ASPECT_RATIOS.include?(aspect_ratio)
           aspect_ratio
         else
--- a/lib/active_storage_validations/aspect_ratio_validator.rb
+++ b/lib/active_storage_validations/aspect_ratio_validator.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require_relative 'shared/asv_active_storageable'
-require_relative 'shared/asv_analyzable'
 require_relative 'shared/asv_attachable'
 require_relative 'shared/asv_errorable'
 require_relative 'shared/asv_optionable'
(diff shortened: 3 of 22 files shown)


mohammednasser-32:
Hey @Mth0158, the tests are failing for one case `minitest (3.2, rails_7_1, mini_magick)`
However it is passing locally when running `BUNDLE_GEMFILE=gemfiles/rails_7_1.gemfile bundle exec rake test`, using ruby 3.2 and changing [metadata](https://github.com/igorkasyanchuk/active_storage_validations/blob/master/lib/active_storage_validations/metadata.rb#L45) to use `DEFAULT_IMAGE_PROCESSOR` (mini_magick)
Do you know how can I reproduce the error to solve it?

Mth0158:
Hi @mohammednasser-32,
I'll correct the test once I am finished refactoring the `metadata.rb` file, so no worries, on that point :)

mohammednasser-32:
@Mth0158 yes right, sorry for the mess..I thought we need to make it similar to content_type (only `content_type_invalid` is there)
I pushed another approach

Mth0158:
No worries, let me know when it's ready for reviewing :)

Mth0158:
Hi @mohammednasser-32,
I was quite busy last week, I'll probably have time to do your review this week, I'll let you know once it's reviewed :)

mohammednasser-32:
@Mth0158 just a friendly reminder whenever you have the time 🙏😄

README.md
```diff
@@ -204,6 +204,7 @@ en:
       aspect_ratio_not_landscape: "must be a landscape image"
       aspect_ratio_is_not: "must have an aspect ratio of %{aspect_ratio}"
       image_not_processable: "is not a valid image"
+      aspect_ratio_invalid: "has invalid aspect ratio"
```
Mth0158, line 207: `has an invalid aspect ratio`

config/locales/da.yml
```diff
@@ -30,3 +30,4 @@ da:
       aspect_ratio_not_landscape: "skal være et landskabsbillede"
       aspect_ratio_is_not: "skal have et størrelsesforhold på %{aspect_ratio}"
       image_not_processable: "er ikke et gyldigt billede"
+      aspect_ratio_invalid: "har et ugyldigt billedformat"
```
Mth0158, line 33: update translations accordingly

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -1,118 +1,144 @@
...
+    include ASVActiveStorageable
+    include ASVAttachable
+    include ASVErrorable
+    include ASVOptionable
+    include ASVSymbolizable
+
+    AVAILABLE_CHECKS = %i[with in].freeze
+    NAMED_ASPECT_RATIOS = %i[square portrait landscape].freeze
+    ASPECT_RATIO_REGEX = /is_([1-9]\d*)_([1-9]\d*)/.freeze
+    PRECISION = 3
```
Mth0158, line 20: why did you remove .freeze?

lib/active_storage_validations/aspect_ratio_validator.rb
Mth0158, line 1: I am not a big fan on how you did your code. Method names are way to vague so it's really difficult to understand what the code is doing, maybe try to update names and change the overall logic a bit to make it clearer?

test/validators/aspect_ratio_validator_test.rb
```diff
@@ -158,21 +158,54 @@
...
+            it { is_expected_to_have_error_message('aspect_ratio_is_not', error_options: error_options) }
             it { is_expected_to_have_error_options(error_options) }
           end
         end
       end
     end
 
-    describe "Edge cases" do
-      describe "when the passed file is not a valid image" do
+    describe ':in' do
```
Mth0158, line 174: Test is ok, thanks :)

Mth0158 (changes requested):
Thanks for the PR,
Overall it is a nice work, however I think your code could be more readable in the validator, it's hard to follow what is happening, I think it's mostly due to your method naming. Maybe try to spend some time improving this? The rest is ok 👍

lib/active_storage_validations/aspect_ratio_validator.rb
mohammednasser-32: rubocop removed it since integer is an immutable object..restored it

test/validators/aspect_ratio_validator_test.rb
```diff
@@ -89,16 +89,16 @@
...
       # validates :with_named_landscape, aspect_ratio: :landscape
       # validates :with_named_square_proc, aspect_ratio: -> (record) { :square }
       # validates :with_named_portrait_proc, aspect_ratio: -> (record) { :portrait }
       # validates :with_named_landscape_proc, aspect_ratio: -> (record) { :landscape }
-      %w(value proc).each do |value_type|
-        describe "named aspect_ratio" do
-          %i(square portrait landscape).each do |named_aspect_ratio|
+      %w[value proc].each do |value_type|
+        describe 'named aspect_ratio' do
+          %i[square portrait landscape].each do |named_aspect_ratio|
```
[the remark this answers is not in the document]
mohammednasser-32: yes, should I remove them?

mohammednasser-32:
@Mth0158 Thank you for the review :pray: looking at the code again and you are absolutely right :sweat_smile: 
did some refactoring and renaming [here](https://github.com/igorkasyanchuk/active_storage_validations/pull/294/commits/9ec5d833395d79c0e1e59652849b89d1aa60c041)..please let me know if more changes are needed, thank you

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -1,118 +1,150 @@
...
+    end
+
+    private
+
+    def is_valid?(record, attribute, attachable, metadata)
+      flat_options = set_flat_options(record)
+
+      return if image_metadata_missing?(record, attribute, attachable, flat_options, metadata)
+
+      aspect_ratios = aspect_ratios(flat_options).compact
```
Mth0158, line 48: As mentioned in the previous review, the code should, IMO, be more explicit.
Here we are talking about allowed aspect ratios, using `aspect_ratios` is very vague and could be interpreted different ways. I suggest naming this method `authorized_aspect_ratios`. Same for the variable.

Reading the code is way easier afterwards :)

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -1,118 +1,150 @@
...
+      errors = add_errors(aspect_ratios, metadata)
```
Mth0158, line 49: Same here, you named for method `add_errors`, but we already have a method doing this (`add_error(record, attribute, error, **errors_options)`).
So, at first, reading your code I am thinking it's doing the same, ie adding the error to the record, but in reality it's not adding any error, it's building possible error message keys.

I was first suggesting naming your method `build_error_keys`. But then, reading the rest of the code, I got the real purpose of your method => validating the fact that the attachable aspect ratio at least passes 1 @allowed_aspect_ratios.

Maybe it could be a better idea to do things like the `content_type_validator` is doing:
```ruby
    def authorized_aspect_ratio?(record, attribute, attachable, metadata)
      attachable_aspect_ratio_is_authorized = @authorized_aspect_ratios.any? do |authorized_aspect_ratio|
        case authorized_aspect_ratio
        when :square then valid_square_aspect_ratio?(metadata)
        ... same for all named aspect ratio
        when ASPECT_RATIO_REGEX then valid_regex_aspect_ratio?(authorized_aspect_ratio, metadata)
        end
      end

      return true if attachable_aspect_ratio_is_authorized

      ... set the error according the current aspect ratio
    end
```

lib/active_storage_validations/aspect_ratio_validator.rb
Mth0158: You could even pass the value to an instance variable called `@authorized_aspect_ratios` to use it without needing it as a parameter

lib/active_storage_validations/aspect_ratio_validator.rb
Mth0158: and that's it, you do not need the reste of the `is_valid?` method anymore, the logic is now in `authorized_aspect_ratio?`

test/validators/aspect_ratio_validator_test.rb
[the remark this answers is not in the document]
Mth0158: Yes they are polluting the code review :/ I'll add rubocop rules in the v2 of the gem

mohammednasser-32:
Thanks again for the review! comments incorporated [here](https://github.com/igorkasyanchuk/active_storage_validations/pull/294/commits/4acaa5873b8907e1c0fc2166eddd5cbe21f0e746)

mohammednasser-32:
Thanks again for the reveiw! added the changes and rebased

Mth0158:
@mohammednasser-32 better adding new correction commits rather than rebasing the whole thing, it makes the review process quite difficult because I have to start over every time :/
Not rebasing would also not make the review comments "Outdated" which is easier to review :)

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -1,120 +1,149 @@
...
+    def check_validity!
+      ensure_at_least_one_validator_option
+      ensure_aspect_ratio_validity
+    end
+
+    def validate_each(record, attribute, _value)
+      return if no_attachments?(record, attribute)
+
+      validate_changed_files_from_metadata(record, attribute)
+    end
```
Mth0158, line 41: Once again, I think that you should read the `content_type_validator`. Here you could do:

```ruby
      @authorized_aspect_ratios = authorized_aspect_ratios_from_options(record)
      return if @authorized_aspect_ratios.empty?
```

That will run the operation once. You are setting `@authorized_aspect_ratios` on every aspect_ratio validation which is not necessary.

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -1,120 +1,149 @@
...
+
+    private
+
+    def is_valid?(record, attribute, attachable, metadata)
+      flat_options = set_flat_options(record)
+
+      return if image_metadata_missing?(record, attribute, attachable, flat_options, metadata)
+
+      @authorized_aspect_ratios = authorized_aspect_ratios(flat_options).compact
```
Mth0158, line 50: To be removed with 1st comment

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -1,120 +1,149 @@
...
+      
+      return true if authorized_aspect_ratio?(record, attribute, attachable, metadata)
+
+      errors_options = initialize_error_options(options, attachable)
+      errors_options[:aspect_ratio] = string_aspect_ratios(flat_options)
+      add_error(record, attribute, aspect_ratio_error_mapping, **errors_options)
+      false
```
Mth0158, line 57: As previously stated, I would suggest doing the same as `content_type_validator` for the sake of clarity.

```ruby
    def is_valid?(record, attribute, attachable, metadata)
      authorized_aspect_ratio?(record, attribute, attachable, metadata)
    end
```

And then do the below logic in the `authorized_aspect_ratio?` method.

lib/active_storage_validations/aspect_ratio_validator.rb
```diff
@@ -1,120 +1,149 @@
...
+        when :square then valid_square_aspect_ratio?(metadata)
+        when :portrait then valid_portrait_aspect_ratio?(metadata)
+        when :landscape then valid_landscape_aspect_ratio?(metadata)
+        when ASPECT_RATIO_REGEX then valid_regex_aspect_ratio?(authorized_aspect_ratio, metadata)
+        end
+      end
+    end
+
+    def aspect_ratio_error_mapping
+      return :aspect_ratio_invalid unless @authorized_aspect_ratios.length == 1
```
Mth0158, line 72: `@authorized_aspect_ratios.length.one?`

Mth0158 (changes requested):
Thanks @mohammednasser-32 for the revamp. But I think there is still room for improvement, I think after this one we are good to go to merge though ;)

mohammednasser-32:
Thank you again and sorry for how messy this turned :sweat_smile:
appreciate your review and comments...changes applied

lib/active_storage_validations/aspect_ratio_validator.rb
mohammednasser-32: I believe `one?` is used for blocks and arrays not integers, used `@authorized_aspect_ratios.one?` instead

Mth0158:
Hi @mohammednasser-32!

Thanks again for your hard work! The PR is a great addition to the gem ✌️

PR #295, opened 2024-11-14
292 validate content type of failing since 131

Mth0158:


Mth0158:
Put to v2 because it will be a breaking change for those using 'image/jpg' content type in a `in` option

Mth0158:
new MR with base to v2

Issue #254, opened 2024-07-19
Refactor the `metadata.rb` using built-in `ActiveStorage::Analyzer::ImageAnalyzer::ImageMagick/Vips.new(blob).metadata`


Mth0158:
Reading Rails documentation to implement a feature for this gem, I came across the following ActiveStorage class: [https://github.com/rails/rails/blob/55c4adeb36eff229972eecbb53723c1b80393091/activestorage/lib/active_storage/analyzer/image_analyzer.rb](https://github.com/rails/rails/blob/55c4adeb36eff229972eecbb53723c1b80393091/activestorage/lib/active_storage/analyzer/image_analyzer.rb)

It looks like it both supports `MiniMagick` and `Vips` since June 2021. The `metadata.rb` file however was built a few years before maybe explaining its presence?

We could rely on `ActiveStorage::Analyzer::ImageAnalyzer`s to do this job, therefore decreasing the weight + associated unnecessary complexity of this gem?

@igorkasyanchuk @gr8bit what do you think of it?

Issue #255, opened 2024-07-19
Handle video media


Mth0158:
We could add a few validators for audio & video files. We would need something to analyse such files and ... tada! There is already one built-in Rails around FFMPEG:
- video: [https://api.rubyonrails.org/v7.1.3.4/classes/ActiveStorage/Analyzer/VideoAnalyzer.html](https://api.rubyonrails.org/v7.1.3.4/classes/ActiveStorage/Analyzer/VideoAnalyzer.html)

We would then just have to build the validations on top on the retrieved metadata, that wouldn't be such a big issue I guess :) A bit of work yes but nothing too hard.

@igorkasyanchuk & @gr8bit what do you think?

Mth0158:
Merged in #310

Issue #199, opened 2023-09-08
Processable image validation seems not working in all cases


floganz:
Hi, 

I've noticed that `processable_image` validation works kind of strange. It's possible to create any invalid image file (like create .txt file and then change extension to .png) and upload it even if that validation enabled in the model.

[Here](https://github.com/floganz/active_storage_validations) is test application I've created to show the issue.
There's both instruction how to reproduce it locally via interface and test that indicated the same behaviour. Sample images also added to that repo.

I'm using Vips for image analysis and it looks like all needed for vips libs installed.
Running `vips --vips-config` on my machine next.
<details>
  <summary>Output</summary>

enable debug: false
enable deprecated: true
enable modules: true
enable cplusplus: true
enable RAD load/save: true
enable Analyze7 load/save: true
enable PPM load/save: true
enable GIF load: true
use fftw for FFTs: true
accelerate loops with ORC: true
ICC profile support with lcms: true
zlib: true
text rendering with pangocairo: true
font file support with fontconfig: true
EXIF metadata support with libexif: true
JPEG load/save with libjpeg: true
JXL load/save with libjxl: true (dynamic module: true)
JPEG2000 load/save with OpenJPEG: true
PNG load/save with libspng: true
PNG load/save with libpng: false
selected quantisation package: imagequant
TIFF load/save with libtiff: true
image pyramid save with libgsf: true
HEIC/AVIF load/save with libheif: true (dynamic module: true)
WebP load/save with libwebp: true
PDF load with PDFium: false
PDF load with poppler-glib: true (dynamic module: true)
SVG load with librsvg: true
EXR load with OpenEXR: true
OpenSlide load: true (dynamic module: true)
Matlab load with libmatio: true
NIfTI load/save with niftiio: false
FITS load/save with cfitsio: true
GIF save with cgif: true
selected Magick package: MagickCore (dynamic module: true)
Magick API version: magick7
Magick load: true
Magick save: true

</details>

Mth0158:
Thanks @floganz for this issue.
It's a bit late, but I'll try to investigate it in the coming days. Do you have any update on that issue btw?

floganz:
I've debugged gem a bit and discovered track trace for this issue and possible solution. However I'm lacking context why it was made in that way, maybe you know why. @Mth0158 

[...]

PR #299, opened 2024-11-24
[Analyzer] Refacto our image analyzers to further expand the gem (#254)

Mth0158:


Mth0158:
Self note: ensure https://github.com/igorkasyanchuk/active_storage_validations/issues/199 is fixed with this?

Mth0158:
Self note: ensure https://github.com/igorkasyanchuk/active_storage_validations/issues/255 is handled with this?

lib/active_storage_validations/analyzer/image_analyzer/image_magick.rb
```diff
@@ -0,0 +1,46 @@
...
+  # This analyzer relies on the third-party {MiniMagick}[https://github.com/minimagick/minimagick] gem.
+  # MiniMagick requires the {ImageMagick}[http://www.imagemagick.org] system library.
+  # This is the default Rails image analyzer.
+  class Analyzer::ImageAnalyzer::ImageMagick < Analyzer::ImageAnalyzer
+
+    private
+
+    def read_image
+      begin
+        require "mini_magick"
```
igorkasyanchuk, line 13: question if we need to require it always here
can we just check if defined?(MiniMagick)?

and maybe raise the error even sooner, in the initializer (check when the app boots if possible)?

Mth0158: I think we do not need to require it every time. That's, indeed, not necessary, plus it has performance benefits.

I am updating the code with a `unless defined?(MiniMagick)` (same for Vips), as it was done previouslu by the `metadata.rb` file (I should have ketp it).

For the second part, we could raise the error at the app boot, but I think this will enforce the user to have both `imagemagick` and `ruby-vips` gem in their `Gemfile` (which we do not want). Plus, if the user does not use validators based on the image analyzers, it's not necessary to have either `imagemagick` or `ruby-vips` gem. Therefore I think it's better to leave it this way? What do you think?

Mth0158: Using `Object.const_defined?(:Vips)` is better than `unless defined?(Vips)` because we might not have loaded the lib yet, therefore it might not be defined yet. Whereas `Object.const_defined?(:Vips)` is about presence in the code.

igorkasyanchuk: I'm fine with bject.const_defined?(:Vips)

I'm also fine if it just raises an exception and we will have something about it in the docs. Not the most user friendly but this should be expected. Bu this option is least preferable.

Mth0158: @igorkasyanchuk I finally opted for `defined?(MiniMagick)`. It is the best option you are right. Let me know if the PR is okay with you now :)

igorkasyanchuk (approved):

PR #326, opened 2024-12-13
Require marcel

ngan:
Saw this error in CI and realized nothing was requiring `marcel`. Things might have been working because ActiveStorage might have loaded it before this gem. But in general, we should require where we use it.

```
/usr/local/bundle/ruby/3.3.0/gems/active_storage_validations-1.3.4/lib/active_storage_validations/marcel_extensor.rb:3:in `<top (required)>': uninitialized constant Marcel (NameError)
  |  
  | Marcel::MimeType.extend "application/x-rar-compressed", parents: %(application/x-rar)
  | ^^^^^^
  | Did you mean?  Arel
  | from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:75:in `require'
  | from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:75:in `block (2 levels) in replace_require'
  | from /usr/local/bundle/ruby/3.3.0/gems/active_storage_validations-1.3.4/lib/active_storage_validations.rb:17:in `<top (required)>'
  | from <internal:/usr/local/lib/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
  | from <internal:/usr/local/lib/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
  | from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:75:in `block (2 levels) in replace_require'
  | from /usr/local/lib/ruby/site_ruby/3.3.0/bundler/runtime.rb:60:in `block (2 levels) in require'
  | from /usr/local/lib/ruby/site_ruby/3.3.0/bundler/runtime.rb:55:in `each'
  | from /usr/local/lib/ruby/site_ruby/3.3.0/bundler/runtime.rb:55:in `block in require'
  | from /usr/local/lib/ruby/site_ruby/3.3.0/bundler/runtime.rb:44:in `each'
  | from /usr/local/lib/ruby/site_ruby/3.3.0/bundler/runtime.rb:44:in `require'
  | from /usr/local/lib/ruby/site_ruby/3.3.0/bundler.rb:212:in `require'
  | from /usr/local/bundle/ruby/3.3.0/gems/sidekiq-ent-7.1.2/bin/sidekiqswarm:40:in `<top (required)>'
  | from bin/sidekiqswarm:27:in `load'
  | from bin/sidekiqswarm:27:in `<main>'
 ```

### DIFF
--- a/lib/active_storage_validations/marcel_extensor.rb
+++ b/lib/active_storage_validations/marcel_extensor.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "marcel"
+
 Marcel::MimeType.extend "application/x-rar-compressed", parents: %(application/x-rar)
 Marcel::MimeType.extend "audio/x-hx-aac-adts", parents: %(audio/x-aac)
 Marcel::MimeType.extend "audio/x-m4a", parents: %(audio/mp4)

### CODE REVIEW

File: lib/active_storage_validations/marcel_extensor.rb [hunk 1]
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "marcel"
```
Mth0158, line 3: `unless defined?(Marcel)`

Mth0158 (changes requested):
Hi @ngan,
Thanks for reporting this issue ! That's a good point indeed.
To not load the lib if already loaded, which could cause loading / perf issues, I suggest adding the below.

lib/active_storage_validations/marcel_extensor.rb
ngan: Hey @Mth0158 — this isn’t a normal pattern in Ruby. If you call require on the same path that has already been loaded, Ruby will no-op very quickly. Probably more quickly that checking if a constant is defined since it’s a hash lookup vs a constant lookup.

ngan: That said, this is your project and if you think that’s better, I’m happy to do it. If you look in rails or other prominent Ruby code based, I’d be surprised if you see anyone do this.

Mth0158: @ngan you are absolutely right, I was not aware of the no-op mecanism (https://ruby-doc.org/3.3.6/Kernel.html#method-i-require).

@igorkasyanchuk that what we were discussing in the other PR, might be a good idea to remove the `unless defined?(MiniMagick/Vips)` in our code?

igorkasyanchuk: Not a blocker from my side, but defined works faster. Or I'm missing something?

```ruby
require "benchmark"

def require_check
  require "marcel"
end

def defined_check
  defined?(Marcel)
end

Benchmark.bm do |x|
  x.report("require") { 1000.times { require_check } }
  x.report("defined") { 1000.times { defined_check } }
end

```

```
ruby app.rb
             user     system      total        real
require  0.025928   0.019764   0.045692 (  0.045691)
defined  0.000067   0.000000   0.000067 (  0.000068)
```

Mth0158: Mmm I would rather do the below, but the `defined_check` is still better

```ruby
require "benchmark"

require "marcel"

def require_check
  require "marcel"
end

def defined_check
  require "marcel" unless defined?(Marcel)
end

Benchmark.bm do |x|
  x.report("require") { 1000.times { require_check } }
  x.report("defined") { 1000.times { defined_check } }
end
```

```sh
             user     system      total        real
require  0.000462   0.000007   0.000469 (  0.000468)
defined  0.000042   0.000000   0.000042 (  0.000046)
```

ngan: I haven't benchmarked it myself, but I think you'd want to bench mark with `marcel` already loaded such that all calls to `require` will no-op, and calls to `defined?(Marcel)` will return the constant, instead of `nil`. I also imagine it depends on the size of the app / running process since you'd have a lot more constants loaded. But again, I'm not sure. I've just never seen the practice in my 18 years of working with Ruby 😅. Maybe this is a good question for the Ruby community? https://www.ruby-lang.org/en/community/

ngan: Ah yeah, looks like I was writing my comment on a stale page. But yeah, @Mth0158 is a better benchmark. I image `defined?(Marcel)` might get _slightly_ slower for real-world apps. It looks like we're now in micro-optimization territory. In which case, I would argue that our objective here is to "load the marcel library" and not "check if the constant Marcel exists." Since...`Marcel` is just a constant that could be defined by anything/anywhere...even a monkey-patch.

Anyways, this was such a fun conversation -- thanks guys! Now I'm really curious what Ruby-core would think/say.

Mth0158: I am really curious about it 🤔. It is all about the perf of calling `require "marcel"` vs calling `defined?(Marcel)`. In the benchmark I have done we see that `defined?(Marcel)` is a 10x faster, but it now in the context of an app that will have many constants defined.

I guess that, making the assumption that `require` and `defined?` perform the check at the same speed, it would be faster to use `defined?` because they are probably less than 10 constant definitions per file (therefore per required file). But that's a lot hypothetical.

Probably that not seing this `defined?` pattern is a sign that's not the most optimised move?

ngan: Not sure if you're in Nate's Rails Performance Slack, but I posted the question there: https://railsperf.slack.com/archives/C0QKHT5LP/p1734109484198909

Mth0158: Thanks @ngan, curious to read the answer :).
I can't join the Slack because it's restricted to some organisations, or you have to be invited by an admin. I would gladly join the slack channel, I am really interesting in theses questions :)

image

ngan: I've asked the Admin to extend you an invite. See you on the other side.

ngan: Until then, one member mentioned this:
https://github.com/igorkasyanchuk/active_storage_validations/blob/1b2d5dc1353276f0c4fa59f48d549aaf20a10b46/lib/active_storage_validations/analyzer/image_analyzer/image_magick.rb#L12-L17

This is somewhat dangerous because:
> It's possible someone might open up the MiniMagic namespace to do like MiniMagic::CustomThing without actually loading the library (it's possible to do that with Modules... not great, but possible)

bensheldon: 👋🏻 💖 The insight I want to share is that:

- `defined?` checks whether a constant exists
- `require` loads a file once, and tracks what has been loaded already via `$LOADED_FEATURES`.

There is some overlap between files and constants, but in most real-world applications, there is deviation too (sometimes reasonable, sometimes unreasonable). As a gem, I would recommend resiliency over correctness, and just use `require`.

PR #327, opened 2024-12-13
Refactor require-ing for supported analyzers

ngan:
From the discussion in https://github.com/igorkasyanchuk/active_storage_validations/pull/326, I thought it'd be a fun exercise to see what it would look like to extract support detection. This PR does the follow:
- Adds a new abstract `supported?` method that allows analyzers to do whatever to determine whether or not it's supported. This usually means loading the library and catching the load error.
- The support list is held (and cached) in a class attribute (Hash) for quick, subsequent lookups.
- This also changes `require "whatever" unless defined?(Whatever)` to just `require "whatever"` since it safeguards against things that happen to define `Whatever` without loading the library/gem.

### DIFF
--- a/lib/active_storage_validations/analyzer/image_analyzer/vips.rb
+++ b/lib/active_storage_validations/analyzer/image_analyzer/vips.rb
@@ -8,13 +8,6 @@ module ActiveStorageValidations
     private
 
     def read_image
-      begin
-        require "vips" unless defined?(::Vips)
-      rescue LoadError
-        logger.info "Skipping image analysis because the ruby-vips gem isn't installed"
-        return {}
-      end
-
       Tempfile.create(binmode: true) do |tempfile|
         begin
           if image(tempfile)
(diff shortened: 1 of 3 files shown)


lib/active_storage_validations/analyzer/image_analyzer/vips.rb
```diff
@@ -47,6 +40,13 @@ def image_from_path(path)
       end
     end
 
+    def supported?
+      require "vips"
+      true
+    rescue LoadError
+      false
```
Mth0158, line 47: Why did you remove the logger.info "..." line? Plus you left in mini_magick analyzer

Mth0158 (changes requested):
Hi again @ngan 
Thanks for this PR, that solves yesterday's point! It's ok to me (except the logger point ?).

I'll also run it through @igorkasyanchuk since he was part of the discussion :) @igorkasyanchuk let us know what you think of this

lib/active_storage_validations/analyzer/image_analyzer/vips.rb
ngan: Whoops! Accidental 😅

Mth0158 (approved):

PR #342, opened 2025-01-02
334 analyzer refacto the content type spoof into an analyzer

Mth0158:


Mth0158:
Tests not passing because of:
```ruby
    def set_attachable_cached_values(attachable)
      @attachable_content_type = disable_spoofing_protection? ? attachable.blob.content_type : attachable_content_type_rails_like(attachable)
      @attachable_filename = disable_spoofing_protection? ? attachable.blob.filename.to_s : attachable_filename(attachable).to_s
    end
```

These lines are further changed in https://github.com/igorkasyanchuk/active_storage_validations/pull/341 so we consider it ok.

PR #345, opened 2025-01-27
Fix wrong Japanese translation

OdenTakashi:
Thank you for developing such a awesome gem. 👍

## Description
I corrected the parts of the Japanese translation that were grammatically incorrect, and revised the variable names that were in Japanese so they are now specified in English.


Mth0158:
Hi @OdenTakashi,
Thanks for helping on the internationalisation side 💪🏼. I'll trust you on the translations since I do not speak Japanese.

PR #400, opened 2025-10-01
Make PDF analyzer handle decimal dimensions

tekin:
PDFs are often generated based on print sizes in inches and centimeters. This can result in decimal pixel or point dimensions. This change updates the PdfAnalyzer to return the correct integer dimensions for such PDFs.


Mth0158:
Hi @tekin 
This looks ok to me, thanks a lot for the PR, it will be released in the coming days :)

PR #403, opened 2025-11-12
fix: correct typo in UK locale content type error messages

miry:
Extend the test helper for error validation to cover all available_locales.
Update locale files accordingly to ensure all tests pass.

One of the problems:
Fixed a typo in the `content_type_invalid` and `content_type_spoofed` error messages for the UK locale. Added a simple test to validate error messages across `available_locales`.

```
I18n::MissingInterpolationArgument: missing interpolation argument :autorized_human_content_types in "має недійсний тип вмісту (дозволені типи вмісту %{autorized_human_content_types})"
```

Sidecar: I have enabled github actions for PRs.


Mth0158:
Thanks a lot for the PR @miry 
Indeed there were some issues with the I18n translations, thanks for adding the related tests. Maybe you could add the I18n for the `is_expected_to_include_error_message` method inside `test/validators/support/validator_helpers.rb`? So that every `is_expected_to_include_error_message` test is doing the check? :)

miry:
I’ve updated the helper to validate error messages across all available locales.
During the update, I noticed that allow_blank relies on locale messages from rails-i18n, so for those cases, I’m only checking the EN locale.

Some error messages also include file sizes with localized byte units, which would require a more sophisticated matcher. Since the current matcher is static and expects English translations, I limited those checks to the EN locale for now.

@Mth0158 pls review

PR #420, opened 2026-04-13
409 do not run heavy validators if size validator failed breaking

### LINKED ISSUE #409: Do not run heavy validators if `size` validator failed (BREAKING)
## Expected behaviour
If the `size` validator has been set for an attribute, and this attribute also have heavy validators (e.g. `dimension`, ...), do not run them if size validation fails.

This will prevent unnecessary heavy operations, plus mitigate the risk of a malicious user uploading a 10 GB file that will get open by our validators.

This is a small breaking change, we change the gem behaviour, but it's for the best and probably not so breaking?

Mth0158:


Mth0158:
This could be integrated one day, but not for now, changing the public API is not something we are leaning into until we have real demand for it.

PR #422, opened 2026-06-08
Fix validation of multiple new attachables with nil blob ids

tmnb:
## Summary
- Fix `CreateMany` attachable deduplication so persisted blobs are deduplicated by `blob.id`, while newly attached unsaved files remain distinct.
- Add a shared regression test covering multiple new files where one file is valid and another is invalid.
- Preserve the intended duplicate signed_id behavior from #421 for persisted blobs.

## Test plan
- `BUNDLE_GEMFILE=gemfiles/rails_8_1.gemfile bundle exec ruby -Itest test/validators/content_type_validator_test.rb -n "/multiple new files/"`
- Ran the same `-n "/multiple new files/"` filter across the validator files that include the shared ASVAttachable behavior.


Mth0158:
Thanks @tmnb, it will be released in 3.0.6 soon :)

